PR #1, opened 2016-12-16
s/OpenCL/EGL/

1ace:
Looks like a simple copy/paste mistake :)


oddhack:
Yeah, I'm doing this for about 15 registries at once. Thanks.

PR #3, opened 2017-01-26
Fix EGL_KHR_context_flush_control path

1ace:
Follow up to cd7c18b163566cd4e3693df6572a00b0635c0997

Note: I couldn't find `regproc.tcl`, so this is untested.

### DIFF
--- a/registry.tcl
+++ b/registry.tcl
@@ -524,7 +524,7 @@ extension EGL_KHR_no_config_context {
 extension EGL_KHR_context_flush_control {
     number      102
     flags       public
-    filename    ../gl/extensions/KHR/context_flush_control.txt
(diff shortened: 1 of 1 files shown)


oddhack:
This should already be fixed in master; I pushed cd7c18b163566cd4e3693df6572a00b0635c0997 yesterday. What's the problem you're seeing? If you're not still seeing it in the current khronos.org/registry/EGL, please close this.

oddhack:
Oh, I see. registry.tcl isn't used for generating anything now. It just has the extension numbers in it. I should probably remove it and use index.php, or turn it into a Python script along the lines of the new GL registry or something. No harm in accepting this but as a general matter I probably won't keep those paths up to date.

PR #6, opened 2017-02-08
genheaders.py: make it usable as a module

groleo:
Signed-off-by: Adrian Negreanu <groleo@gmail.com>


stonesthrow:
Assigned OddHack to implement if he sees the benefit.

PR #7, opened 2017-02-28
Add 'EXT' suffix to 'EGL_METADATA_SCALING' in EXT_surface_SMPTE2086_metadata

weiwliu:


oddhack:
Merging this, but just to note, it was never added to api/egl.xml and so is not getting included in eglext.h.

weiwliu:
That's fine. I was going to add it to egl.xml, but then I saw Courtney already did it in his pending change.

PR #9, opened 2017-03-01
Add EGL_EXT_surface_CTA861_3_EXT extension

courtney-g:
Adds two enums that go with the SMPTE 2086 metadata
that allow application to set metadata described in the
CTA861.3 extension.
[ChangeLog] Add EGL_EXT_surface_CTA861_3_EXT extension

### DIFF
--- a/registry.tcl
+++ b/registry.tcl
@@ -604,4 +604,9 @@ extension EGL_EXT_compositor {
     flags       public
     filename    extensions/EXT/EGL_EXT_compositor.txt
 }
+extension EGL_EXT_surface_CTA861_3_metadta {
+    number      117
+    flags       public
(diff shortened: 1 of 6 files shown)


registry.tcl
```diff
@@ -604,4 +604,9 @@ extension EGL_EXT_compositor {
     flags       public
     filename    extensions/EXT/EGL_EXT_compositor.txt
 }
+extension EGL_EXT_surface_CTA861_3_metadta {
+    number      117
+    flags       public
+    filename    extensions/EXT/EGL_EXT_surface_CTA861_3_metadata.txt
+}
 # Next free extension number: 117
```
oddhack: For future reference, please make sure to update this number (now 118).

I might try a different representation of the extension number database soon, though. This is a holdover from the old setup and I like what I changed to for GL better.

PR #13, opened 2017-05-16
API: Fix pointers in EXT_image_dma_buf_import_modifiers

fooishbar:
The spec erroneously listed the pointer component as part of the name,
rather than the type.


oddhack:
@1ace I fixed the other issue you identified and pushed that as well. There are no header changes corresponding to this, fortunately.

PR #14, opened 2017-06-13
Add EGL_EXT_image_implicit_sync_control extension

fooishbar:
New EXT, should be shipped in Mesa shortly, backing from ARM to ship as EXT

### DIFF
--- a/api/EGL/egl.h
+++ b/api/EGL/egl.h
@@ -38,7 +38,7 @@ extern "C" {
 
 #include <EGL/eglplatform.h>
 
-/* Generated on date 20170621 */
+/* Generated on date 20170622 */
 
 /* Generated C header for:
  * API: egl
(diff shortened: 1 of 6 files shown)


fooishbar:
@oddhack I can't seem to assign this to you, presumably since I'm not marked as a maintainer of this repo. Not that it bothers me, but is a doc bug as the process doc on the other wiki says to explicitly assign it to you.

oddhack:
There seems to be no way for non-KhronosGroup members to assign a PR/Issue, which is sad. In the meantime another extension got registered and you will need to re-do the PR to accomodate. Sorry about that.

fooishbar:
@oddhack No problem, I've rebased now and renumbered to 120. The 'next free number' didn't get updated in registry.tcl for colorspace_scrgb BTW, so this commit jumps it from 119 (clashes with scrgb) to 121 (correct after adding this as 120).

PR #17, opened 2017-06-17
Restore revision and date stamping

1ace:
At some point (transition from svn to git?), the revision and date have stopped being stamped in the generated headers.

### DIFF
--- a/api/reg.py
+++ b/api/reg.py
@@ -16,6 +16,7 @@
 
 import io,os,re,string,sys
 from lxml import etree
+import subprocess
 
 def write(*args, **kwargs):
     file = kwargs.pop('file', sys.stdout)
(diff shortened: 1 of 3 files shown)


api/reg.py
```diff
@@ -524,7 +525,11 @@ def beginFile(self, genOpts):
         #
         # User-supplied prefix text, if any (list of strings)
         if (genOpts.prefixText):
+            git_rev = subprocess.check_output(['git', 'rev-parse', '--short=10', 'HEAD']).decode('utf-8').strip()
```
1ace: If `git` is not available, the python script fails, and the file isn't generated. If `git` fails, the script goes ahead and uses whatever output git gave (probably nothing, with this command) to perform the substitution.
I can change this if you want, just tell me what behaviour you'd like :)

1ace:
> As long as this is being fixed, I'd like to see it changed to just say something like 'Github revision SHA'. The date is imbedded in the following macro anyway.

One is the date when the file was generated, while the other is the date when its source was last modified. Those are two different informations, but if we don't care about the latter I'm happy to remove it.

api/reg.py
oddhack: I'd prefer it generate an empty string, or "Unknown", or something of that nature, rather than failing.

1ace:
In case you missed the revision, this new version does this substitution:

    -** Khronos $Revision$ on $Date$
    +** Khronos $Git commit SHA1: a732b061e7 $ on $Git commit date: 2017-06-17 23:27:53 +0100 $

With both `Revision` and `Date` set to `unknown` if git is unavailable or fails.

PR #29, opened 2017-10-13
Fix use of invalid enum

courtney-g:
Accidentally used EGL_INVALID_VALUE which doesn't actually
exist. EGL_FALSE used instead.


weiwliu:
@courtney-g I don't think EGL_FALSE is a valid error code though. The function should return EGL_FALSE when it fails, and also generate maybe a EGL_BAD_PARAMETER error.

courtney-g:
Dang, you're right. I confused return value with error generation.
I'll make another patch.

courtney-g:
New MR: https://github.com/KhronosGroup/EGL-Registry/pull/31

PR #27, opened 2017-10-11
Add EGL_ANDROID_get_native_client_buffer

critsec:

PR #25, opened 2017-10-03
EGL_ANDROID_get_frame_timestamps

briandersn:
This is live in Android O now, so I am upstreaming.
It defines previously allocated enum values.

PR #30, opened 2017-10-14
Regenerate headers for EGL_ANDROID_get_native_client_buffer

klausw:
This was missing from https://github.com/KhronosGroup/EGL-Registry/pull/27,
@critsec said they had run make, but the resulting change wasn't included
in commit a9b3522c3b26600f0aaca489ac10d1f99c255f39 as merged.

### DIFF
--- a/api/EGL/egl.h
+++ b/api/EGL/egl.h
@@ -33,12 +33,12 @@ extern "C" {
 ** used to make the header, and the header can be found at
 **   http://www.khronos.org/registry/egl
 **
-** Khronos $Git commit SHA1: 65416b63de $ on $Git commit date: 2017-10-03 14:37:16 -0700 $
+** Khronos $Git commit SHA1: 6480009b57 $ on $Git commit date: 2017-10-11 22:51:16 -0700 $
 */
 
 #include <EGL/eglplatform.h>
 
-/* Generated on date 20171004 */
+/* Generated on date 20171013 */
 
 /* Generated C header for:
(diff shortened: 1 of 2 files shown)


critsec:
I assumed Jon would regenerate those when merged, like he does for Vulkan changes (I guess that's not visible to non-Khronos-mebers?). Including them in pull requests tends to lead to unnecessary merge conflicts.

Jon, if you want people to include these in extension PRs, I can add that to the README.md -- I just followed the process described there.

klausw:
For context, @briandersn apparently regenerated headers himself in PR #25, and @kenrussell had asked me to do so for this change also in https://chromium-review.googlesource.com/c/chromium/src/+/679961 . 

The toplevel README just says "Verify that the EGL headers regenerate properly after applying your XML changes" and doesn't ask for the changes to be included specifically. I guess either approach works, having them regenerated makes the PRs a bit more self contained at the cost of extra merge conflicts.

kenrussell:
Thanks Klaus for taking care of this.

klausw:
FYI, this will add a global "struct AHardwareBuffer" declaration to this include file, and as far as I can tell this will happen unconditionally on all platforms, not just Android.

Is that a problem? It seems as if the current global names are prefixed with "EGL" or "khronos".

oddhack:
Thanks. I just forgot to do it in this instance. Conflicts here will be uncommon since the repo isn't very active so people can go either way.

PR #31, opened 2017-10-16
WIP: Fix use of invalid enum

courtney-g:
Replace EGL_FALSE from v7 with EGL_BAD_ATTRIBUTE.
Mixed return value and error code in v7.

### DIFF
--- a/extensions/EXT/EGL_EXT_surface_SMPTE2086_metadata.txt
+++ b/extensions/EXT/EGL_EXT_surface_SMPTE2086_metadata.txt
@@ -13,6 +13,7 @@ Contact
 Contributors
 
     Tom Cooksey
+    Courtney Goeltzenleuchter
     Mathias Heyer
     Lauri Hyvarinen
     James Jones
@@ -30,7 +31,7 @@ Status
 
 Version
 
-    Version 7 - Oct 13, 2017
+    Version 8 - Oct 16, 2017
 
 Number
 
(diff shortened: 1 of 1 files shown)


extensions/EXT/EGL_EXT_surface_SMPTE2086_metadata.txt
```diff
@@ -239,3 +240,8 @@ Revision History
     Version 7, 2017/10/13
       - Rename EGL_INVALID_VALUE (which doesn't exist) to EGL_FALSE
 
+    Version 8, 2017/10/16
+      - Fix v7 change to use EGL_BAD_ATTRIBUTE as the error code
```
weiwliu: This is not the same error as on line 132...
I'd actually prefer BAD_PARAMETER over BAD_ATTRIBUTE, because the attributes are valid, just their values are out of range.
There's also another EGL_FALSE that needs to be corrected on line 217.

oddhack: @courtney-g I'll just hold off a bit on merging this while you deal with this comment. Let me know when you're ready (maybe we should adopt the "WIP" prefix on github, too?)

courtney-g: I thought I was good to go, it had been several days.
I'll add WIP until I get okay from @weiwliu

weiwliu: @courtney-g there's one more EGL_FALSE in "issue #1". Change should be good after fixing that.

weiwliu:
Looks good

courtney-g:
Thanks @weiwliu!
@oddhack - ready to merge.

PR #32, opened 2017-11-14
Const-qualify attrib_list in eglStreamConsumerGLTextureExternalAttrib…

jdashg:
…sNV.


jdashg:
We ran into this while using the extension for sampling from d3d11 textures in ANGLE.

oddhack:
@jdashg please add the matching change to egl.xml, and I'll accept this. Thanks.

jdashg:
@oddhack I updated the xml and regenerated the headers.

jdashg:
Rebased on master.

PR #33, opened 2017-11-21
Add EGL_NV_context_priority_realtime extension

sashinde:
This extension allows an EGLContext to be created with one extra priority level in addition to three levels that are part of EGL_IMG_context_priority extension. 

This new level has extra privileges that are not available to other three levels.

### DIFF
--- a/api/egl.xml
+++ b/api/egl.xml
@@ -791,7 +791,9 @@
         <enum value="0x3350" name="EGL_GL_COLORSPACE_SCRGB_LINEAR_EXT"/>
         <enum value="0x3351" name="EGL_GL_COLORSPACE_SCRGB_EXT"/>
         <enum value="0x3352" name="EGL_TRACK_REFERENCES_KHR"/>
-            <unused start="0x3353" end="0x335F"/>
+            <unused start="0x3353" end="0x3356"/>
+        <enum value="0x3357" name="EGL_CONTEXT_PRIORITY_REALTIME_NV"/>
+            <unused start="0x3358" end="0x335F"/>
         <enum value="0x3360" name="EGL_CTA861_3_MAX_CONTENT_LIGHT_LEVEL_EXT"/>
         <enum value="0x3361" name="EGL_CTA861_3_MAX_FRAME_AVERAGE_LEVEL_EXT"/>
         <enum value="0x3362" name="EGL_GL_COLORSPACE_DISPLAY_P3_LINEAR_EXT"/>
@@ -2819,6 +2821,11 @@
             <require>
                 <enum name="EGL_COVERAGE_BUFFERS_NV"/>
                 <enum name="EGL_COVERAGE_SAMPLES_NV"/>
+            </require>
+        </extension>
+        <extension name="EGL_NV_context_priority_realtime" supported="egl">
+            <require>
+                <enum name="EGL_CONTEXT_PRIORITY_REALTIME_NV"/>
             </require>
         </extension>
         <extension name="EGL_NV_coverage_sample_resolve" supported="egl">
--- a/extensions/NV/EGL_NV_context_priority_realtime.txt
+++ b/extensions/NV/EGL_NV_context_priority_realtime.txt
@@ -1,0 +1,195 @@
+Name
+
+    NV_context_priority_realtime
+
+Name Strings
+
+    EGL_NV_context_priority_realtime
+
+Contributors
+
+    Sandeep Shinde
+    Kirill Artamonov
+    Sami Kiminki
+    Donghan Ryu
+    Daniel Koch
+    Mark Kilgard
+
+Contacts
+
+    Sandeep Shinde, NVIDIA (sashinde 'at' nvidia 'dot' com)
+
+Status
+
+    Complete
+
+Version
+
+    Version 4 - November 21, 2017
+
+Number
+
+    EGL Extension #124
+
+Dependencies
+
+    Requires EGL 1.0.
+
+    Requires EGL_IMG_context_priority
+
+    This extension is written against the wording of the EGL 1.5
+    Specification - August 27, 2014 (but may be implemented against earlier
+    versions).
+
+Overview
+
+    This extension allows an EGLContext to be created with one extra
+    priority level in addition to three priority levels that are part of
+    EGL_IMG_context_priority extension.
+
+    This new level has extra privileges that are not available to other three
+    levels. Some of the privileges may include:
+    - Allow realtime priority to only few contexts
+    - Allow realtime priority only to trusted applications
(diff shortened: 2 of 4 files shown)


extensions/NV/EGL_NV_context_priority_realtime.txt
```diff
@@ -0,0 +1,194 @@
...
+    7)  Can a context created with a realtime priority hint that is
+        in fact given a realtime priority, subsequently find that realtime
+        priority revoked and, if revoked, can it be restored?
+
+        RESOLVED: No, once a context is created with specific priority level, the
+        priority will not change for lifetime of the context. This means there will
+        not be revoking or restoring of realtime priority to already created context.
+
+    8)  The attrib_list for eglCreateContext could list the
+        EGL_CONTEXT_PRIORITY_REALTIME_NV attribute (or some other
```
1ace: You mean the `EGL_CONTEXT_PRIORITY_LEVEL_IMG` attribute.

On the issue itself, I feel like giving several priority at once is an API usage error, and should either return an error or be undefined/implementation-defined behaviour.

`or some other attribute` should not be in this extension spec, as it shouldn't change the behaviour of all other specs and extensions

api/egl.xml
```diff
@@ -791,7 +791,9 @@
         <enum value="0x3350" name="EGL_GL_COLORSPACE_SCRGB_LINEAR_EXT"/>
         <enum value="0x3351" name="EGL_GL_COLORSPACE_SCRGB_EXT"/>
         <enum value="0x3352" name="EGL_TRACK_REFERENCES_KHR"/>
-            <unused start="0x3353" end="0x335F"/>
+            <unused start="0x3353" end="0x3356"/>
+        <enum value="0x3357" name="EGL_CONTEXT_PRIORITY_REALTIME_NV"/>
```
1ace: Any reason to add it in the middle of this free range?
No objection, it's your block, I'm just curious :)

extensions/NV/EGL_NV_context_priority_realtime.txt
```diff
@@ -0,0 +1,194 @@
...
+    None
+
+New Procedures and Functions
+
+    None
+
+New Tokens
+
+    New attribute values accepted in the <attrib_list> argument
+    of eglCreateContext:
```
1ace: I would specify the attribute this applies to:
> New attribute values accepted for the EGL_CONTEXT_PRIORITY_LEVEL_IMG attribute in the <attrib_list> argument of eglCreateContext

extensions/NV/EGL_NV_context_priority_realtime.txt
```diff
@@ -0,0 +1,194 @@
...
+        such applications have to be registered in advance with the OS;
+        unpriviledged third party app cannot acquire this capability.
+
+        This restriction exists so arbitrary applications do not starve or
+        otherwise compromise the interactivity of the system overall.
+
+    6)  In practice how many realtime priority contexts can exist in a system?
+
+        RESOLVED: Only one realtime priority context should be active at one moment
+        to get best performance.
```
1ace: This doesn't really answer the question; the lack of answer kind of implies there is no limit, but I think you should answer explicitly the question you're asking

api/egl.xml
sashinde: Enum Values from 0x3353-0x3356 are being used for one of our internal extensions, which we've not made public yet, so just using next free available from our internal range.

extensions/NV/EGL_NV_context_priority_realtime.txt
sashinde: Right, I'll update the spec.

extensions/NV/EGL_NV_context_priority_realtime.txt
sashinde: I think I should rephrase the question: "In practice how many realtime priority contexts can exist in a system to get best performance of it?". 

Because it's just a guideline on how many should exists for best performance.

oddhack:
@sashinde this all seems fine to me. Let me know when you're done dealing with other feedback on the PR and I'll accept it.

I do prefer that extensions just update the XML and leave updating the header to me. It's not a problem most of the time but it helps avoid collisions when multiple PRs are in flight.

extensions/NV/EGL_NV_context_priority_realtime.txt
sashinde: >>  I feel like giving several priority at once is an API usage error, and should either return an error or be undefined/implementation-defined behaviour.

Yes and no, it's not a usual behavior to pass same attribute twice with different values but it's not an error too. We've chosen not to return error but use the last value in the array as the one that app intended to be selected.

As far as I know, there is no error returned for other attributes also which are already part of EGL spec, if multiple values are specified for same attribute (multiple attrib-value pairs for same attrib) then there is no language in Spec which says which one should be chosen.

sashinde:
>> I do prefer that extensions just update the XML and leave updating the header to me. It's not a problem most of the time but it helps avoid collisions when multiple PRs are in flight.

I'm going to re-generate pull request with some changes as I just commented on queries from Eric. So do you want me to generate this new request without header file changes and just XML change + api spec?

Thanks for the reviews.

oddhack:
> I'm going to re-generate pull request with some changes as I just commented on queries from Eric. So do you want me to generate this new request without header file changes and just XML change + api spec?

That would be great - thanks.

sashinde:
I've uploaded new commit with changes discussed.

extensions/NV/EGL_NV_context_priority_realtime.txt
dgkoch: @sashinde you still didn't fix the issue - what @1ace was pointing out is that the attribute is EGL_CONTEXT_PRIORITY_LEVEL_IMG which could be listed multiple times (EGL_CONTEXT_PRIORITY_REALTIME_NV could be a value associated with the attribute).

sashinde: @dgkoch  Yes right, I'll just update the patch with below as Issue 8:

    8)  The attrib_list for eglCreateContext could list the
        EGL_CONTEXT_PRIORITY_LEVEL_IMG attribute multiple
        times with different valid values. What happens in this case?

dgkoch: that sounds good.  I also agree with @1ace that the behaviour in this case is undefined and what you wrote is just what our implementation happens to do currently.

sashinde: @dgkoch Done, updated Issue and Answer both in latest patch.

PR #35, opened 2017-11-22
Change shebangs to use python2. (PEP 394)

jdashg:


oddhack:
This might happen to work now, but that could change without warning in the future as compatibility with an 8 year old release isn't a priority. If you must use python 2, you can override PYTHON in the Makefile.

At some point we might shift EGL and GL over to use the more currently maintained XML schema and scripts being developed in Vulkan and other projects, and those definitely require python3.

jdashg:
Ok, thanks!

PR #36, opened 2017-11-23
EGL_MESA_drm_image: fix a couple typos and add a cursor use bit

1ace:
This pulls in the Mesa changes from Kristian Høgsberg (@krh) and Nicolas Kaiser (@nikai3d)

https://cgit.freedesktop.org/mesa/mesa/commit/?id=e5169e9615e8391ea369415b356168717b8f7be0
https://cgit.freedesktop.org/mesa/mesa/commit/?id=ae5776c41f12515bb73c07ee2a0aed56cdd1a1ef

### DIFF
--- a/api/egl.xml
+++ b/api/egl.xml
@@ -149,6 +149,7 @@
     <enums namespace="EGLDRMBufferUseMESAMask" type="bitmask" comment="EGL_DRM_BUFFER_USE_MESA bits">
         <enum value="0x00000001" name="EGL_DRM_BUFFER_USE_SCANOUT_MESA"/>
         <enum value="0x00000002" name="EGL_DRM_BUFFER_USE_SHARE_MESA"/>
+        <enum value="0x00000004" name="EGL_DRM_BUFFER_USE_CURSOR_MESA"/>
     </enums>
 
     <!-- Should be shared with GL, but aren't aren't since the
--- a/extensions/MESA/EGL_MESA_drm_image.txt
+++ b/extensions/MESA/EGL_MESA_drm_image.txt
@@ -24,7 +24,7 @@ Number
 
 Dependencies
 
-    Reguires EGL 1.4 or later.  This extension is written against the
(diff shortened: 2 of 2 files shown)


extensions/MESA/EGL_MESA_drm_image.txt
```diff
@@ -66,6 +66,7 @@ New Tokens
 
         EGL_DRM_BUFFER_USE_SCANOUT_MESA		0x0001
         EGL_DRM_BUFFER_USE_SHARE_MESA		0x0002
+        EGL_DRM_BUFFER_USE_CURSOR_MESA	0x0004
```
oddhack: We normally name bitfield enums 'EGL_name_BIT[_vendor]', but that ship has probably sailed with this extension. However, the new bit does need to be added to egl.xml as part of this PR.

1ace: You're right, I completely forgot that.
I should let you regenerate the headers though, right?

oddhack: Yes, I'll regenerate them when this is accepted. Thanks!

api/egl.xml
```diff
@@ -149,6 +149,7 @@
     <enums namespace="EGLDRMBufferUseMESAMask" type="bitmask" comment="EGL_DRM_BUFFER_USE_MESA bits">
         <enum value="0x00000001" name="EGL_DRM_BUFFER_USE_SCANOUT_MESA"/>
         <enum value="0x00000002" name="EGL_DRM_BUFFER_USE_SHARE_MESA"/>
+        <enum value="0x00000004" name="EGL_DRM_BUFFER_USE_CURSOR_MESA"/>
```
oddhack: This needed to be added to the \<extension> block as well - I did this and regenerated the header after merging this.

1ace: Indeed; should've grepped. Hopefully now I'll get it right in one go next time :)
Thanks for merging all my PR :+1:

PR #39, opened 2018-02-27
Add extension EGL_EXT_image_gl_colorspace

tweenk:
This is a minimal extension that allows one to override the color space of an EGLImage. This is mainly intended to support creating sRGB images from AHardwareBuffers on Android. The extension was already discussed and approved by Khronos OpenGL ES and EGL working groups.


oddhack:
Please add the extension to egl.xml, including its interface, so header generation will work for it as well. I'll accept it once that's done.

tweenk:
Oops, forgot about that. Added missing content to XML file and regenerated headers.

PR #48, opened 2018-04-12
Reserve enum values for ANGLE

courtney-g:
Need more enums for ongoing ANGLE development.
bug: 7


api/egl.xml
```diff
@@ -887,6 +887,9 @@
         <enum value="0x3471" name="EGL_IMPORT_IMPLICIT_SYNC_EXT"/>
         <enum value="0x3472" name="EGL_IMPORT_EXPLICIT_SYNC_EXT"/>
     </enums>
+    <enums namespace="EGL" start="0x3480" end="0x3F8F" vendor="ANGLE" comment="Reserved for Courtney Goeltzenleuchter - ANGLE (gitlab EGL bug 7)">
```
oddhack: I am hoping you don't actually need values 0x3480-0x3F8F, but some smaller quantity? See below.

api/egl.xml
```diff
@@ -887,6 +887,9 @@
...
+            <unused start="0x3480" end="0x3FFF"/>
```
oddhack: If you only need 16 values as line 908 below suggests, then the "end" values here and immediately above should be "0x348F".

courtney-g:
Updated to only request 16 enums.

PR #67, opened 2018-11-14
man: fix eglGetProcAddress() function declaration

1ace:
Rendered, it looked like this:

    void (* eglGetProcAddress)()(char const * procname);

That declaration was invalid, but my closest interpretation would've been that `eglGetProcAddress` was a pointer to a function that returns a function that takes a procname and doesn't return anything. Not really accurate.

This changes it to look like this instead:.

    void * eglGetProcAddress(char const * procname);

A function that takes a procname and returns a `void*` (to be cast into the actual function).

### DIFF
--- a/sdk/docs/man/docbook4/eglGetProcAddress.xml
+++ b/sdk/docs/man/docbook4/eglGetProcAddress.xml
@@ -23,7 +23,7 @@
         <title>C Specification</title>
         <funcsynopsis>
             <funcprototype>
-                <funcdef>void (* <function>eglGetProcAddress</function>)()</funcdef>
+                <funcdef>void * <function>eglGetProcAddress</function></funcdef>
                 <paramdef>char const * <parameter>procname</parameter></paramdef>
             </funcprototype>
         </funcsynopsis>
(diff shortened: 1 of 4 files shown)


stonesthrow:
+1

oddhack:
The way we define this in the headers is

typedef void (*__eglMustCastToProperFunctionPointerType)(void);  
EGLAPI __eglMustCastToProperFunctionPointerType EGLAPIENTRY
       eglGetProcAddress(const char *procname);

which avoids the complex declarations; the spec says

void (*eglGetProcAddress(const char *procname))(void);

which is a more compact way of saying the same thing. I don't think there's good reason to have yet a third way in the refpages (also, I don't believe ISO C requires function/data pointers to be the same size, or convertible, although the arcana and lore around that are vast and deep).

1ace:
> I don't think there's good reason to have yet a third way in the refpages

My reasoning was that the ref pages are meant to be read by humans, not by machines, so readability matters ¯\_(ツ)_/¯

Would you prefer if I changed this PR to only remove the invalid `()` in the middle?

oddhack:
I suggest using the spec prototype. Removing the '()' from 'void (* eglGetProcAddress)()(char const * procname);' gives valid C but it isn't the correct definition of the command.

stonesthrow:
Are we going to progress this further?

oddhack:
@stonesthrow I'd prefer #72.

stonesthrow:
Very good. move to #72

Issue #69, opened 2018-12-04
Unable to specify non-linear Display-P3 with no HW transfer function


courtney-g:
There are cases where an application may want to apply the sRGB transfer function as part of it's shader computations and would like to use the non-linear Display-P3 color space. However, the current EGL_EXT_gl_colorspace_display_p3 will always use the HW sRGB support for 8-bit pixel formats.

oddhack:
I think #71 should have closed this - if not please reopen and note what remains TBD.

PR #71, opened 2018-12-12
Add passthrough version of Display P3 colorspace

courtney-g:
Discovered that there was no way for an application to specify
they want to use the non-linear Display P3 colorspace but do not
want the HW to apply the transfer function, instead the application
is responsible for applying the transfer function.

Bug #69

### DIFF
--- a/api/EGL/egl.h
+++ b/api/EGL/egl.h
@@ -33,12 +33,12 @@ extern "C" {
 ** used to make the header, and the header can be found at
 **   http://www.khronos.org/registry/egl
 **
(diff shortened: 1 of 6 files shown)


stonesthrow:
@courtney-g assign to Jon and say you are ready to merge if ready.

stonesthrow:
Wiki has process to follow. Work that as close as possible.
https://github.com/KhronosGroup/EGL-Registry/wiki

oddhack:
All good!

PR #72, opened 2018-12-19
Alternate fix for #67 - use the correct prototype as defined in the spec

oddhack:
The Docbook XML syntax I came up with for this function pointer declaration is wonky, but the rendered result comes out correctly.


stonesthrow:
@oddhack seems fine to me

oddhack:
Going ahead and merging / closing #67. If there's a remaining problem, please open a new one.

PR #78, opened 2019-03-12
EXT_image_dma_buf_import_modifiers: Correct no-modifier-token behaviour

fooishbar:
Zero is a valid and explicit modifier token, meaning that the image is
pitch-linear, i.e. no tiling or compression is applied.

When no modifier is passed for dmabuf import, the implementation may
assume the image is linear, but is not required to do so. Many drivers
which supported tiling formats before modifiers (or which do not yet
support modifiers) will determine the layout via other mechanisms, e.g.
metadata held by the kernel.

Fix the wording for the case where the user does not explicitly pass
modifiers to state that the layout will be determined by the
implementation, rather than stating the layout must be linear.

Reported by @olvaffe

### DIFF
--- a/extensions/EXT/EGL_EXT_image_dma_buf_import_modifiers.txt
+++ b/extensions/EXT/EGL_EXT_image_dma_buf_import_modifiers.txt
@@ -24,7 +24,7 @@ Status
 
 Version
 
-    Version 4, October 20, 2016
+    Version 5, March 12, 2019
 
 Number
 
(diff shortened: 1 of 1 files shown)


fooishbar:
Ouch, apparently the force-push caused @olvaffe's comment to disappear. Sorry about that. :(

Copied below:

> Maybe it is best to change
> 
> `If neither of the two attributes are given, the format modifier is assumed to be zero.`
> to
>
> `If neither of the two attributes are given, or if the given format modifier is DRM_FORMAT_MOD_INVALID, the effective format modifier is implementation-defined.`
> Apparently, there are already clients (e.g., Mutter) that use DRM_FORMAT_MOD_INVALID to avoid getting linear modifier (i.e., zero).
>
> On the other hand, the clients are violating
>
> ```
> Implementations are required to validate the full combination of base
> format and per-plane modifiers, and reject any combination which is not
> explicitly supported.
> ```
> We have the right to break them. But I think it is probably better not to do that.
>
> We should also clarify that eglQueryDmaBufModifiersEXT must not include DRM_FORMAT_MOD_INVALID.

stonesthrow:
As an EXT, are there other supporters, besides Collabora, that can thumbs up this change so it can be recommended to Jon for merge? Or is this a Collabora specific extension?

1ace:
Intel approves of this clarification :+1:

fooishbar:
@stonesthrow It's implemented for most Mesa drivers, and I believe the Arm proprietary driver as well.

stonesthrow:
thumbs-up

PR #80, opened 2019-03-20
Add EGL_NV_n_buffer extension.

santanu-thangaraj:


santanu-thangaraj:
@oddhack @dkartch-nv , please take a look at this.

api/egl.xml
```diff
@@ -2924,6 +2926,12 @@
                 <enum name="EGL_CUDA_DEVICE_NV"/>
             </require>
         </extension>
+        <extension name="EGL_NV_n_buffer" supported="egl">
```
oddhack: According to the spec, there are two different extensions in the single document, each supporting a single enum. If that's how the driver exposes support for this functionality, then there should be two \<extension> blocks named accordingly - AFAICT there's no actual extension named "EGL_NV_n_buffer". It's fine if the extension .txt document is named that but not the header interfaces.

oddhack: The new parts look fine, but I believe this \<extension> block should be removed.

santanu-thangaraj: Missed out during merge step, now fixed with latest patches.

oddhack:
This looks fine now. If it's ready to merge from your perspective, let me know and I'll hit the switch.

1ace:
It would probably be best for those 4 fixup commits to be squashed into a single commit for this PR before merging it, right?

oddhack:
Github can squash during the merge process, so I think there's no need for that.

oddhack:
@santanu-thangaraj so I'm just waiting for final signoff from you before merging this - please reply to this comment.

santanu-thangaraj:
@oddhack , we can go ahead and merge this. Thanks.

PR #81, opened 2019-04-25
Add KHRONOS_STATIC to allow static linking on Windows.

tweenk:
I have encountered multiple situations where it is desirable to statically link against a software implementation of EGL on Windows. Add the preprocessor constant KHRONOS_STATIC that disables the annotation of EGL entry points as DLL-imported.


stonesthrow:
This addition seems fine. +1

PR #82, opened 2019-05-01
Add support for Emscripten in EGL/eglplatform.h.

tweenk:
The Emscripten toolchain for WebAssembly defines the native EGL handle
types as integers in its custom eglplatform.h header. Make the Khronos
version compatible with it.

### DIFF
--- a/api/EGL/eglplatform.h
+++ b/api/EGL/eglplatform.h
@@ -77,6 +77,12 @@ typedef HDC     EGLNativeDisplayType;
 typedef HBITMAP EGLNativePixmapType;
 typedef HWND    EGLNativeWindowType;
 
+#elif defined(__EMSCRIPTEN__)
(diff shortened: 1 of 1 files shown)


stonesthrow:
The sections of this file are for each window system so there is a definition for window system native types. Emscripten is a tool, not window system specific, correct? Is using Emscripted truly orthogonal to any other Window System/OS? As is, can I be using Emscripten with say X11 or Win32 or ...? The defined types are fine, if this is specific to a separate window system.

tweenk:
Code compiled by Emscripten runs as a script or a WebAssembly module in a web browser and cannot directly interact with native platorm APIs, so Emscripten is effectively a distinct OS platform with a separate window system.

stonesthrow:
OK. +1.

PR #84, opened 2019-05-17
Update egl.xml with missing QCOM private extensions

tate-hornbeck:
Fill in used, but undocumented QCOM private extensions


oddhack:
This is great - thanks for the updates. Hopefully the underlying extensions can be documents in the registry at some point, too.

PR #85, opened 2019-05-20
Add EGL_NV_stream_origin extension

mvicomoya:
This extension provides new attributes to allow EGL stream users to query frame orientation and whether it is handled automatically by the producer or consumer endpoints so that clients are not required to take further actions.


mvicomoya:
@oddhack, could you help with this? If not, who could?

PR #87, opened 2019-07-17
Add prototype guards to EGL header.

null77:
This allows the header to be used in applications that define their
own prototypes. Such as for entry point loaders. The new header is
backwards compatible with the previous standard header.

@ianelliottus @chrisforbes any feedback? Could ANGLE upstream this change?

### DIFF
--- a/api/EGL/egl.h
+++ b/api/EGL/egl.h
@@ -33,12 +33,16 @@ extern "C" {
 ** used to make the header, and the header can be found at
 **   http://www.khronos.org/registry/egl
 **
-** Khronos $Git commit SHA1: 836d10e67a $ on $Git commit date: 2019-05-17 06:29:03 -0700 $
+** Khronos $Git commit SHA1: d738c39fe2 $ on $Git commit date: 2019-06-14 16:17:31 -0600 $
 */
 
 #include <EGL/eglplatform.h>
 
-/* Generated on date 20190520 */
+#ifndef EGL_EGL_PROTOTYPES
+#define EGL_EGL_PROTOTYPES 1
+#endif
+
+/* Generated on date 20190716 */
 
 /* Generated C header for:
  * API: egl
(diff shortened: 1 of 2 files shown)


stonesthrow:
I'd like to ask for a few members to do a review and a compile check for all the different platforms. Then we can give a Oddhack a thumbs up of confidence to proceed with the merge. I assume null77 you have done a compile check with ANGLE and Android?

stonesthrow:
@cubanismo @shadazar please review and give it a compile check.

null77:
@stonesthrow we currently use this version of egl.h with ANGLE on Windows, Mac, Linux, Android and Fuchsia.

1ace:
@stonesthrow: I'm running this through Intel's CI right now to make sure this works on everything we care about, but I don't expect any issues. Results will be on this page when that's done (in about 45 minutes I expect):
https://mesa-ci.01.org/eric_engestrom/builds/117/group/63a9f0ea7bb98050796b649e85481845

@null77:
> This allows the header to be used in applications that define their own prototypes. Such as for entry point loaders.

Do you have an example of what such an application would look like with your change?

shadazar:
When I generated the header files I noticed that in eglext.h the code uses:
#ifdef EGL_EGLEXT_PROTOTYPES
but in egl.h it uses
#if EGL_EGL_PROTOTYPES

So from what I see this will remove the prototypes from the eglext.h file by default and include them in egl.h by default. I would think the behavior for these should be the same. Maybe the eglext.h file should be using #if EGL_EGL_PROTOTYPES.

It also seems odd to me the typedefs are outside of the a #else statement. I would think the idea would be to either have the prototypes or the function type definitions, but not both.

null77:
> @null77:
> 
> > This allows the header to be used in applications that define their own prototypes. Such as for entry point loaders.
> 
> Do you have an example of what such an application would look like with your change?

Sure. See one of ANGLE's EGL loaders here:

https://github.com/google/angle/blob/master/util/egl_loader_autogen.cpp
https://github.com/google/angle/blob/master/util/egl_loader_autogen.h

> When I generated the header files I noticed that in eglext.h the code uses:
> #ifdef EGL_EGLEXT_PROTOTYPES
> but in egl.h it uses
> #if EGL_EGL_PROTOTYPES
> 
> So from what I see this will remove the prototypes from the eglext.h file by default and include them in egl.h by default. I would think the behavior for these should be the same. Maybe the eglext.h file should be using #if EGL_EGL_PROTOTYPES.

This PR uses the same generator as the other OpenGL headers. See:

https://github.com/KhronosGroup/OpenGL-Registry/blob/master/api/GLES3/gl32.h
https://github.com/KhronosGroup/OpenGL-Registry/blob/master/api/GLES2/gl2ext.h

I think what you're proposing is a change that would affect all OpenGL specs and is probably outside the scope of this PR. Does it sound okay to skip that for now?

> It also seems odd to me the typedefs are outside of the a #else statement. I would think the idea would be to either have the prototypes or the function type definitions, but not both.

Same observations here.

shadazar:
> 
> > When I generated the header files I noticed that in eglext.h the code uses:
> > #ifdef EGL_EGLEXT_PROTOTYPES
> > but in egl.h it uses
> > #if EGL_EGL_PROTOTYPES
> > So from what I see this will remove the prototypes from the eglext.h file by default and include them in egl.h by default. I would think the behavior for these should be the same. Maybe the eglext.h file should be using #if EGL_EGL_PROTOTYPES.
> 
> This PR uses the same generator as the other OpenGL headers. It follows the same style as the other GLES and GL headers. See:
> 
> https://github.com/KhronosGroup/OpenGL-Registry/blob/master/api/GLES3/gl32.h
> https://github.com/KhronosGroup/OpenGL-Registry/blob/master/api/GLES2/gl2ext.h
> 
> I think what you're proposing is a change that would affect all OpenGL specs and is probably outside the scope of this PR. Does it sound okay to skip that for now?


@null77 Honestly, I do not think it is okay to skip this for now, either the header files should be setup to use prototypes or setup hide them. Anything less introduces an inconsistency between the header files, and breaks all of my applications that use extensions since they all assume prototypes are available, even the ones in the eglext.h file. Such a change would require that I go to every application and test, and add a #define EGL_EGLEXT_PROTOTYPES before including eglext.h.

null77:
> > > When I generated the header files I noticed that in eglext.h the code uses:
> > > #ifdef EGL_EGLEXT_PROTOTYPES
> > > but in egl.h it uses
> > > #if EGL_EGL_PROTOTYPES
> > > So from what I see this will remove the prototypes from the eglext.h file by default and include them in egl.h by default. I would think the behavior for these should be the same. Maybe the eglext.h file should be using #if EGL_EGL_PROTOTYPES.
> > 
> > 
> > This PR uses the same generator as the other OpenGL headers. It follows the same style as the other GLES and GL headers. See:
> > https://github.com/KhronosGroup/OpenGL-Registry/blob/master/api/GLES3/gl32.h
> > https://github.com/KhronosGroup/OpenGL-Registry/blob/master/api/GLES2/gl2ext.h
> > I think what you're proposing is a change that would affect all OpenGL specs and is probably outside the scope of this PR. Does it sound okay to skip that for now?
> 
> @null77 Honestly, I do not think it is okay to skip this for now, either the header files should be setup to use prototypes or setup hide them. Anything less introduces an inconsistency between the header files, and breaks all of my applications that use extensions since they all assume prototypes are available, even the ones in the eglext.h file. Such a change would require that I go to every application and test, and add a #define EGL_EGLEXT_PROTOTYPES before including eglext.h.

Sorry, can you help me understand your concern? This PR does not change eglext.h. And it makes the function prototypes available by default. So I'm confused as to how this could break your application. It also is consistent with the existing GL header design so I don't understand how this would introduce an inconsistency between header files.

stonesthrow:
Historically, the core functions were statically exported, and the extensions, which might not be implemented were obtained by eglGetProcAddress, but optionally could be exported by the driver, so in that case, an app could set the EGL_EGLEXT_PROTOTYPES and link to the extension functions.

Given the precedent of gl/gles, and given use case, and no impact, I have no objection. 

I am not seeing this yet, that since EGL_EGL_PROTOTYPES is defined at top of egl.h, then how do you undefine this to hide them?

null77:
> I am not seeing this yet, that since EGL_EGL_PROTOTYPES is defined at top of egl.h, then how do you undefine this to hide them?

In our loader we hide them by setting `#define EGL_EGL_PROTOTYPES 0`. Again the same goes for other GL headers, e.g. `#define GL_GLES_PROTOTYPES 0`. The `#ifndef` allows the zero override.

stonesthrow:
Of course. #if checks the statement, not if its defined. OK +1

stonesthrow:
@oddhack , I think this is ready to merge. No issues were raised from a compile check.

null77:
Thanks everyone!

PR #89, opened 2019-08-12
man: Fix a few copy-paste-o's

nathankidd:
A few minor typos I noticed.

I couldn't get the makefile/docbook generation to run so updated the xhtml by hand. :(
I didn't update docbook4 by hand (it no longer is in sync with the .xml/.xhtml).  Not sure if that's intentional or not.


oddhack:
Looks fine. The docbook toolchain is pretty complex, unfortunately.

PR #90, opened 2019-08-14
Add EGL_NO_X11 define for unix platforms that don't want to use X11

tomwardio:
This is equivalent to the already existing USE_OZONE define, used in the Chromium project. The purpose of this define is to be more explicit about it's meaning, making it easier to understand from the client code's perspective.

### DIFF
--- a/api/EGL/eglplatform.h
+++ b/api/EGL/eglplatform.h
@@ -116,6 +116,12 @@ typedef intptr_t EGLNativeDisplayType;
 typedef intptr_t EGLNativePixmapType;
 typedef intptr_t EGLNativeWindowType;
 
+#elif defined(__unix__) && defined(EGL_NO_X11)
+
+typedef void             *EGLNativeDisplayType;
+typedef khronos_uintptr_t EGLNativePixmapType;
+typedef khronos_uintptr_t EGLNativeWindowType;
+
 #elif defined(__unix__) || defined(USE_X11)
 
 /* X11 (tentative)  */


tomwardio:
> Trivial: Without comment on the technical side (I can see reasons for it myself), my inner grammar unit wants to tweak the log message with s/it's/its/, and rewrite the first line to something like 'Support NO_USE_X11 define to avoid including X11' (fits standard length limit, doesn't say "define" twice).

Thanks for suggestions! Have changed log message, hopefully it reads a bit better (and that I updated the PR correctly, not well versed in amending git commits).

1ace:
We need something like this, and I was actually planning on looking at upstreaming something that could replace [our Mesa hack](https://gitlab.freedesktop.org/mesa/mesa/blob/master/include/EGL/eglplatform.h#L121) but hadn't had the time to look at the details yet.

This proposal looks reasonable, but I would prefer we use the pointer types like we have in Mesa rather than plain int types; would that work for you?
That means putting that in an `#elif` with the new typedefs.

It bothers me a bit that you could have `USE_X11` and `NO_USE_X11` together, but I don't really see an alternative, and the precedence of NO overriding USE is the right one in my opinion.

One last nit, I would call it `NO_X11` as I think that's clearer than `NO_USE_X11`.

1ace:
Something like this:
```patch
diff --git a/api/EGL/eglplatform.h b/api/EGL/eglplatform.h
index 29ab288b57eb6ccce047..835bdcba1f440bb460c7 100644
--- a/api/EGL/eglplatform.h
+++ b/api/EGL/eglplatform.h
@@ -116,6 +116,12 @@ typedef intptr_t EGLNativeDisplayType;
 typedef intptr_t EGLNativePixmapType;
 typedef intptr_t EGLNativeWindowType;
 
+#elif defined(__unix__) && defined(NO_X11)
+
+typedef void             *EGLNativeDisplayType;
+typedef khronos_uintptr_t EGLNativePixmapType;
+typedef khronos_uintptr_t EGLNativeWindowType;
+
 #elif defined(__unix__) || defined(USE_X11)
 
 /* X11 (tentative)  */
```

tomwardio:
@1ace thanks for the suggestions!

re: NO_USE_X11, I agree that this is a little unclear (was really just to mirror the existing define), but is there potential for NO_X11 to be used by other packages, potentially causing a conflict? WDYT about "EGL_NO_X11" to make it explicit that this is for disabling X11 only in EGL context?

nathankidd:
Since I'm still on the trivial train:  I would squash those 3 commits to one, since nobody needs the noise.

tomwardio:
Thanks @nathankidd, I hope I've done this correctly!

oddhack:
I use the squash-on-merge option when accepting PRs, so no need to do it yourself if that's the only reason.

1ace:
Looks all good to me (with both my Intel and Mesa hats on) :+1:

tomwardio:
Thanks all. @nathankidd @oddhack is there anything else needed to get this PR accepted?

oddhack:
We should get signoff from the EGL Working Group - @stonesthrow could you bring this up? I added a new 'Approved to Merge' milestone which you could set when the WG signs off, since I rarely attend the meetings any more.

stonesthrow:
PR raised in GL/GLES/EGL agenda today. No objections. I sent email out to EGL membership for review/comment. Wait a couple days for late review/feedback and then proceed with merge. 👍

cubanismo:
:+1:

fooishbar:
Also looks great to me. Would be nice to have the Mesa hack we already have be generic. :+1:

oddhack:
Merged. It might be worth documenting the various macros such as EGL_NO_X11 somewhere, which I don't think we do at present.

PR #95, opened 2019-11-13
Add EGL_WL_bind_wayland_display and EGL_WL_create_wayland_buffer_from_image

fooishbar:
These extensions were developed by the Wayland project several years
ago, but never proposed for inclusion in the primary Khronos repository.

Unfortunately this includes tokens which alias with already-defined
values for the EGL platform extensions. Downstream implementations also
offered non-standard typedefs (e.g. PFNEGLBINDWAYLANDDISPLAYWL instead
of PFNEGLBINDWAYLANDDISPLAYWLPROC) which will have to be accounted for
in those implementations.

### DIFF
--- a/index.php
+++ b/index.php
@@ -339,6 +339,10 @@ include_once("../../assets/static_pages/khr_page_top.php");
 </li>
 <li value=135> <a href="extensions/NV/EGL_NV_stream_dma.txt">EGL_NV_stream_dma</a>
 </li>
+<li value=136> <a href="extensions/WL/EGL_WL_bind_wayland_display.txt">EGL_WL_bind_wayland_display</a>
+</li>
+<li value=137> <a href="extensions/WL/EGL_WL_create_wayland_buffer_from_image.txt">EGL_WL_create_wayland_buffer_from_image</a>
+</li>
 </ol>
 
 <h6> Providing Feedback on the Registry </h6>
--- a/api/EGL/egl.h
+++ b/api/EGL/egl.h
@@ -33,7 +33,7 @@ extern "C" {
 ** used to make the header, and the header can be found at
 **   http://www.khronos.org/registry/egl
 **
-** Khronos $Git commit SHA1: cb927ca98d $ on $Git commit date: 2019-08-08 01:05:38 -0700 $
+** Khronos $Git commit SHA1: 0f3a3945bf $ on $Git commit date: 2019-11-13 13:25:18 +0000 $
 */
 
 #include <EGL/eglplatform.h>
@@ -42,7 +42,7 @@ extern "C" {
 #define EGL_EGL_PROTOTYPES 1
 #endif
 
-/* Generated on date 20190808 */
+/* Generated on date 20191212 */
 
 /* Generated C header for:
  * API: egl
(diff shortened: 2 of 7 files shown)


fooishbar:
These were originally developed by @krh and @bpeel. They have been shipped in Mesa as well as Wayland for quite some time now. We would like to stop doing it downstream and have them included in the primary registry - despite our mistake of plucking an arbitrary unregistered vendor token ('WL' for the Wayland project) and accidentally aliasing enum values with Mesa's platform tokens.

fooishbar:
@oddhack Yeah, that would be incredibly helpful and stop us shipping some bad hacks downstream! Should I do that next to the typedefs like `EGLDEBUGPROCKHR`?

fooishbar:
@oddhack I've replaced the commits in this branch with one that adds typedefs for the legacy entrypoints. This was a little bit hairy, so let me know if it was the right thing to do or if there's another suitable way.

krh:
Thanks @fooishbar.

oddhack:
I think this can be simplified into using either typedefs of the old and new names, or #defines (if you don't mind that approach). Will give it a try if you wish, probably not until tomorrow though.

fooishbar:
> I think this can be simplified into using either typedefs of the old and new names, or #defines (if you don't mind that approach). Will give it a try if you wish, probably not until tomorrow though.

Thanks, I'm fine with anything which actually works, so if you think you have a better approach please feel free. No rush on the time either: it's been 8.5 years already, so I'm sure we can live with another couple of days.

index.php
```diff
@@ -339,6 +339,10 @@
 </li>
 <li value=135> <a href="extensions/NV/EGL_NV_stream_dma.txt">EGL_NV_stream_dma</a>
 </li>
+<li value=136> <a href="extensions/NV/EGL_WL_bind_wayland_display.txt">EGL_WL_bind_wayland_display</a>
+</li>
+<li value=137> <a href="extensions/NV/EGL_WL_create_wayland_buffer_from_image.txt">EGL_WL_create_wayland_buffer_from_image</a>
+</li>
```
1ace: `s/NV/WL/` (the folder in the link URL)

fooishbar:
> No rush on the time either: it's been 8.5 years already, so I'm sure we can live with another couple of days.

I hope I wasn't too literal with the wording here :P

index.php
fooishbar: Fixed, thanks.

oddhack:
Sorry about that. See #97 for an example. It is not elegant but hopefully will suffice. A better job would require out of scope changes to the schema and generator scripts.

fooishbar:
@oddhack Thanks a lot for your help! I've revised this so it uses the `#define` method instead of generating two typedefs (which, in hindsight, would have been treated as distinct types by the compiler).

I tested this with Weston, which uses the old non-`PROC` names, and it built fine.

stonesthrow:
If this WL extension is ready, just mark as "Approved for Merge" and assign to oddhack to have that completed.

fooishbar:
@stonesthrow PR submitters can't approve their own PRs, but I have assigned to @oddhack for merging.

PR #96, opened 2019-12-10
Add support for Fuchsia platform.

digit-google:
This patch only ensures that EGLNative{Display,Pixmap,Window}Type
are defined to adequately-sized generic types. There is no native
EGL support on Fuchsia, which only provides Vulkan APIs.

However, this definition opens the door to porting libraries like
ANGLE or SwiftShader to the platform. For example, see:

  https://swiftshader-review.googlesource.com/c/SwiftShader/+/38488

### DIFF
--- a/api/EGL/eglplatform.h
+++ b/api/EGL/eglplatform.h
@@ -146,6 +146,12 @@ typedef void              *EGLNativeDisplayType;
 typedef khronos_uintptr_t  EGLNativePixmapType;
 typedef khronos_uintptr_t  EGLNativeWindowType;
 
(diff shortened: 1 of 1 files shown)


oddhack:
@pdaniell-nv this seems completely benign, recommend accepting soon.

pdaniell-nv:
+1 from me. @stonesthrow do you want to take a look too?

stonesthrow:
Sorry for delayed response. Looks fine to me. +1

stonesthrow:
Assigned @oddhack to merge.

PR #97, opened 2019-12-12
Example of defining backwards-compatibility names 

oddhack:
for EGL function pointer types in egl.xml. Unfortunately this approach uses #defines
rather than typedefs, due to constraints in the generator scripts, but it should work OK. This example would have to be converted into the right extension / names for the branch underlying #95 - this PR exists only as an example.


stonesthrow:
Only as example - no action here?

oddhack:
Right, just to have a place where they could see the change. Closing.

PR #106, opened 2020-05-22
Don't let KHRONOS_STATIC affect the calling convention

mstorsjo:
When intending to link statically, we should leave out
__declspec(dllimport), but the calling convention (__stdcall) should
still be the same.

This reverts parts of commit 94ba8ee876206364cf45a9bc08b8db5a52cb9543,
making the exception to __stdcall specific to __SCITECH_SNAP__ again
as it was before.

### DIFF
--- a/api/KHR/khrplatform.h
+++ b/api/KHR/khrplatform.h
@@ -119,7 +119,7 @@
  * This follows the return type of the function  and precedes the function
  * name in the function prototype.
  */
-#if defined(_WIN32) && !defined(_WIN32_WCE) && !defined(KHRONOS_STATIC)
+#if defined(_WIN32) && !defined(_WIN32_WCE) && !defined(__SCITECH_SNAP__)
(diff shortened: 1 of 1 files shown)


stonesthrow:
Reference: https://github.com/KhronosGroup/EGL-Registry/pull/81

mstorsjo:
Thanks!

Well, with the previous change, if they applied `KHRONOS_STATIC` consistently both when building libEGL and the caller of it, it shouldn't change anything (it will use a different calling convention but both change in sync).

If libEGL and the caller didn't both use `KHRONOS_STATIC` (but was built for static linking by other means, e.g. the build of libEGL did `-DEGLAPI=`, which happens e.g. when ANGLE is built as part of Qt), it's currently impossible to link against it using `KHRONOS_STATIC`.

So in short, cases that worked before should still work, and it should fix mismatches in other cases.

stonesthrow:
Can be closed.

PR #107, opened 2020-06-18
Remove an unmatched </li> tag.

kbrenneman:
This just removes a superfluous `</li>` tag from index.php.


stonesthrow:
@oddhack this looks trivial.

PR #108, opened 2020-06-30
extensions: Add EGL_EXT_device_query_name.

kbrenneman:
This is a proposal for a simple EGL extension to let an application query a vendor name and device name from an `EGLDeviceEXT` handle.

It just adds `EGL_VENDOR` and a new `EGL_RENDERER_EXT` as valid queries for `eglQueryDeviceStringEXT`.

### DIFF
--- a/registry.tcl
+++ b/registry.tcl
@@ -719,4 +719,9 @@ extension EGL_NV_stream_consumer_eglimage {
     flags       public
     filename    extensions/NV/EGL_NV_stream_consumer_eglimage.txt
 }
-# Next free extension number: 140
+extension EGL_NV_stream_consumer_eglimage {
+    number      140
+    flags       public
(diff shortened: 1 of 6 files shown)


stonesthrow:
@cubanismo - poke James to see if he has feedback.

stonesthrow:
LGTM. +1

registry.tcl
```diff
@@ -719,4 +719,9 @@ extension EGL_NV_stream_consumer_eglimage {
     flags       public
     filename    extensions/NV/EGL_NV_stream_consumer_eglimage.txt
 }
-# Next free extension number: 140
+extension EGL_NV_stream_consumer_eglimage {
+    number      140
```
oddhack, line 723: The extension number here and in the spec text will need to be updated when the PR is actually accepted, up through 141 are used at the moment.

PR #109, opened 2020-07-02
add EGL_EXT_device_query_vendor

1ace:
Firefox [requested](https://gitlab.freedesktop.org/mesa/mesa/-/issues/2976) to have a way to query the device's vendor PCI ID and device PCI ID, so here's an extension that adds this.

Cross-references:
- spec (here)
- [libglvnd ext addition]()
- [Mesa implementation]()
- [implementation test]()

/cc @rmader


extensions/EXT/EGL_EXT_device_query_vendor.txt
```diff
@@ -0,0 +1,77 @@
...
+
+Changes to section 3.2 (Devices)
+
+    Replace the paragraph immediately following the prototype for
+    eglQueryDeviceAttribEXT:
+
+    <attribute> may be either EGL_QUERY_VENDOR_ID_MESA or EGL_QUERY_DEVICE_ID_MESA.
+    On success, EGL_TRUE is returned, and the EGL device's PCI vendor ID or device ID
+    is returned in <value>. If the EGL device is not on a PCI bus, or is not
+    a hardware device at all, EGL_FALSE is returned and <value> is unchanged.
```
anholt, line 68: Should we throw EGL_BAD_ATTRIBUTE, too?  Doesn't EGL generally always throw an error on false returns?  (genuine question, I'm unsure)

Other than that, this sounds great to me.

extensions/EXT/EGL_EXT_device_query_vendor.txt
```diff
@@ -0,0 +1,77 @@
...
+New Tokens
+
+    None.
+
+Changes to section 3.2 (Devices)
+
+    Replace the paragraph immediately following the prototype for
+    eglQueryDeviceAttribEXT:
+
+    <attribute> may be either EGL_QUERY_VENDOR_ID_MESA or EGL_QUERY_DEVICE_ID_MESA.
```
anholt, line 65: Oh, don't you mean `EGL_VENDOR or EGL_DEVICE_EXT` here?

kbrenneman: New attributes would make more sense. `EGL_VENDOR` is currently used to query a string, and `EGL_DEVICE_EXT` always corresponds to an EGLDeviceEXT handle. The PCI vendor and device ID's are both 16-bit integers.

extensions/EXT/EGL_EXT_device_query_vendor.txt
kbrenneman: Every EGLDeviceEXT has its own extension string. If it's not a PCI device, then it should just not advertise this extension at all.

extensions/EXT/EGL_EXT_device_query_vendor.txt
```diff
@@ -0,0 +1,77 @@
+Name
+
+    EXT_device_query_vendor
```
kbrenneman, line 3: "query_vendor" doesn't really describe this extension, since it adds a query for both the vendor and device ID's. Also, "vendor" sometimes means "driver" in GL/EGL.

Maybe something like `EGL_EXT_device_query_pci`?

extensions/EXT/EGL_EXT_device_query_vendor.txt
1ace: I made new attributes originally, but then I decided to use the existing ones as they are unambiguous in the `eglQueryDeviceAttribEXT()` context. I did forget to fix it here, good catch :upside_down_face:

I'm happy to go back to new attributes if that's the consensus :+1:

extensions/EXT/EGL_EXT_device_query_vendor.txt
anholt: I like that solution.

extensions/EXT/EGL_EXT_device_query_vendor.txt
1ace: Good point, I'll rename it :+1:

`EGL_EXT_device_query_pci_id` perhaps?

anholt: Good point.  `EGL_EXT_device_query_pci_id` sounds very clearly descriptive to me.

kbrenneman:
Would it make sense to add queries for the domain/bus/slot ID's as well?

extensions/EXT/EGL_EXT_device_query_vendor.txt
1ace: In mesa, we advertise the same client extensions for all the drivers/devices. I haven't looked at how to change that, but I think I'd prefer if we allowed to advertise this extension on non-PCI devices and have it return `EGL_FALSE`

>  Doesn't EGL generally always throw an error on false returns? (genuine question, I'm unsure)

I can't think of anything off the top of my head (too tired), but I think there are other cases where "false" is not an api error and just a "nope, can't do this" error

extensions/EXT/EGL_EXT_device_query_vendor.txt
1ace: I've fixed it with the re-used attributes for now, but if a second person agrees that we should be adding new enums instead I'll change that everywhere

1ace:
> Would it make sense to add queries for the domain/bus/slot ID's as well?

Why not, but is there a use-case for these?

kbrenneman:
One use I can think of would be to distinguish between multiple identical devices.

extensions/EXT/EGL_EXT_device_query_vendor.txt
1ace: I'll make new enums for `EGL_QUERY_PCI_{VENDOR,DEVICE}_EXT` since I need to make new ones for `EGL_QUERY_PCI_{DOMAIN,BUS,SLOT,FUNCTION}_EXT` anyway

1ace:
It way beyond the problem we're trying to solve here, but it's trivial to implement and conceptually it belongs in the same extension so I'm adding these and re-renaming the spec to `EGL_EXT_device_query_pci` :+1:

extensions/EXT/EGL_EXT_device_query_vendor.txt
1ace: OK, so I was misunderstanding device extensions quite a bit, thanks Kyle for pointing things out to me.

I have looked into the way we do things in Mesa and, while not very memory efficient (we'd waste a few dozen bytes of static memory per device), it's trivial to only expose the extension when the device is on a PCI bus, so let's do that :+1:

I've updated the spec text, let me know if this is ok?

1ace: Marking as resolved, since there isn't any possible error anymore

kbrenneman:
As a minor nitpick, the "The query always succeeds and returns EGL_TRUE, and <value> is set appropriately" line probably isn't necessary, since that's the default assumption for EGL functions.

Other than generic errors (invalid EGLDeviceEXT handle, out of memory, etc), EGL functions list the errors that they can produce.

In this case, it doesn't make much difference, though, since I don't think an implementation could run into any internal errors for those queries. Anything that did go wrong would have gone wrong in eglQueryDevicesEXT instead.

Anyway, change that if you want. The spec looks good to me either way.

1ace:
Thanks for the feedback; you're right, I'll drop the line.

I also was wondering if I should explicitly say that `EGL_QUERY_PCI_VENDOR_ID_EXT` is the vendor ID, etc. or if it's all obvious enough as it is?

kbrenneman:
I'd say it's obvious enough what each attribute is.

stonesthrow:
Change Milestone to "Approved to Merge" when ready and assign to oddhack for Jon to merge.

1ace:
@stonesthrow I don't have permission to change the milestones; could you give me permission, or do it for me? Thanks!

stonesthrow:
@oddHack: Ready to mearge

1ace:
Actually can this wait? @evelikov has requested a change of enum values, I'm waiting on @fooishbar to confirm.

registry.tcl
```diff
@@ -719,4 +719,9 @@ extension EGL_NV_stream_consumer_eglimage {
     flags       public
     filename    extensions/NV/EGL_NV_stream_consumer_eglimage.txt
 }
-# Next free extension number: 140
+extension EGL_EXT_device_query_pci {
+    number      140
```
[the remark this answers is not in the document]
oddhack: @1ace @evelikov @fooishbar where does this extension stand - was your issue about enum values resolved? Would be nice to get it finished and published.

jlHertel:
Hey folks,

I have a new use case for this extension and I'm wondering what people might think about it:
In [adriconf][1] we make use of libdrm in order to enumerate all devices and get their PCI IDs, but ideally I would like to see us getting this directly from EGL. Enumerating devices is already possible under EGL, but getting their PCI Ids isn't.

Sorry for reviving this old thread. If there is a better place to ask about this I kindly ask you to point me there.

[1]: https://gitlab.freedesktop.org/mesa/adriconf

PR #110, opened 2020-08-13
Switch to Apache 2.0 license on headers, and use SPDX license identifiers

oddhack:
This migrates the format of license headers to use SPDX license identifiers, instead of the full Apache license text. It is otherwise a no-op (although there is a small possibility someone would complain about changing eglplatform.h to Apache 2.0, in which case we'd just change it to Apache 2.0 OR MIT).

I probably won't try and do full REUSE license checker compliance on this repository, given the many extension specs we don't own copyrights or licenses on, and can't change, but changing the headers has downstream benefits.

### DIFF
--- a/api/1.0/EGL/egl.h
+++ b/api/1.0/EGL/egl.h
@@ -1,3 +1,6 @@
+/* Copyright 2006-2020 The Khronos Group Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
 /* Sketchy version of egl.h (really only for reserving
  * enumerant values to EGL tokens and sanity checking
  * prototypes).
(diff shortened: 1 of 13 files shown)


stonesthrow:
So long as the license change is OK, the edit looks fine.

oddhack:
> So long as the license change is OK, the edit looks fine.

I don't know any way to establish that, short of publishing it. If there is a problem it would come from a GPL2-licensed project and we do have a mechanism in place to address that now - we just didn't default to the MIT/Apache dual license because using the dual license requires a more formal decision process and notification of the Promoters.

oddhack:
@stonesthrow please give this "Approved" Milestone when you're comfortable, or merge it yourself.

PR #112, opened 2020-08-28
extensions: Add EGL_EXT_platform_xcb.

yshui:
This enables GL applications to be written without any involvement of Xlib.

* Mesa [implementation](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/6474)

This is my first attempt ever at writing an EGL extension, I don't expect to have done everything right the first try. I hope I can get as much help as possible.

@fooishbar @evelikov @1ace suggestions?

### DIFF
--- a/extensions/EXT/EGL_EXT_platform_xcb.txt
+++ b/extensions/EXT/EGL_EXT_platform_xcb.txt
@@ -76,21 +76,25 @@ New Behavior
     To determine if the EGL implementation supports this extension, clients
     should query the EGL_EXTENSIONS string of EGL_NO_DISPLAY.
 
-    This extension defines the same set of behaviors as the X11 platform
-    extension, except Xlib types are replaced with xcb types.
+    This extension defines the same set of behaviors as EGL_EXT_platform_x11,
+    except Xlib types are replaced with xcb types.
 
     To obtain an EGLDisplay backed by an X11 screen, call
     eglGetPlatformDisplayEXT with <platform> set to EGL_PLATFORM_XCB_EXT. The
     <native_display> parameter specifies the X11 display connection to use, and
-    must point to a valid xcb `xcb_connection_t` or be NULL.  If <native_display>
-    is EGL_DEFAULT_DISPLAY, then EGL will create [1] a connection to the default
-    X11 display. The environment variable DISPLAY determines the default X11
-    display as described in the documentation of `xcb_connect`.  The value of
-    attribute EGL_PLATFORM_XCB_SCREEN_EXT specifies the X11 screen to use. If
-    the attribute is omitted from <attrib_list>, then the environment varialbe
-    DISPLAY will be used to select a screen.  Otherwise, the attribute's value
-    must be a valid screen on the display connection. If the attribute's value
-    is not a valid screen, then an EGL_BAD_ATTRIBUTE error is generated.
+    must point to a valid xcb `xcb_connection_t` or be EGL_DEFAULT_DISPLAY.  If
+    <native_display> is EGL_DEFAULT_DISPLAY, then EGL will create [1] a
+    connection to the default X11 display. The environment variable DISPLAY
+    determines the default X11 display, and, unless overridden by the
+    EGL_PLATFORM_XCB_SCREEN_EXT attribute, the default X11 screen - as
+    described in the documentation of `xcb_connect`.  If the environment
+    variable DISPLAY is not present in this case, the result is undefined. The
+    value of attribute EGL_PLATFORM_XCB_SCREEN_EXT specifies the X11 screen to
+    use. If the attribute is omitted from <attrib_list>, and <native_display>
+    is not EGL_DEFAULT_DISPLAY, then screen 0 will be used. Otherwise, the
+    attribute's value must be a valid screen on the display connection. If the
+    attribute's value is not a valid screen, then an EGL_BAD_ATTRIBUTE error is
+    generated.
 
     [fn1] The method by which EGL creates a connection to the default X11
     display is an internal implementation detail. The implementation may use
@@ -106,7 +110,24 @@ New Behavior
 
 Issues
 
-    None so far.
+    1. As xcb_connection_t doesn't carry a screen number, how should a screen be
+       selected in eglGetPlatformDisplayEXT()?
+
+       RESOLVED. The screen will be chosen with the following logic:
+
+         * If EGL_PLATFORM_XCB_SCREEN_EXT is specified, it will always take
+           precedence. Whether <native_display> is EGL_DEFAULT_DISPLAY or not.
+
+         * Otherwise, if <native_display> is not EGL_DEFAULT_DISPLAY, then
+           screen 0 will be used.
+
+         * Otherwise, which is to say <native_display> is EGL_DEFAULT_DISPLAY.
+           Then the DISPLAY environment variable will be used to determine the
+           screen number. If DISPLAY contains a screen number, that will be
+           used; if not, then 0 will be used.
+
+         * If the DISPLAY environment variable is not present when
+           <native_display> is EGL_DEFAULT_DISPLAY, the result will be undefined.
 
 Example Code
 
@@ -352,6 +373,10 @@ Example Code
     }
 
 Revision History
+
+    Version 2, 2020.10.13 (Yuxuan Shui)
+        - Some wording changes
+        - Address the question about screen selection
 
     Version 1, 2020.08.28 (Yuxuan Shui)
         - First draft
--- a/extensions/EXT/EGL_EXT_platform_xcb.txt
+++ b/extensions/EXT/EGL_EXT_platform_xcb.txt
@@ -16,7 +16,7 @@ Contacts
 
 Status
 
-    Draft
+    Complete
 
 Version
 
--- a/extensions/EXT/EGL_EXT_platform_xcb.txt
+++ b/extensions/EXT/EGL_EXT_platform_xcb.txt
@@ -1,0 +1,357 @@
+Name
+
+    EXT_platform_xcb
+
+Name Strings
+
+    EGL_EXT_platform_xcb
+
+Contributors
+
+    Yuxuan Shui <yshuiv7@gmail.com>
+
+Contacts
+
+    Yuxuan Shui <yshuiv7@gmail.com>
+
+Status
+
+    Draft
+
+Version
+
+    Version 1, 2020-08-28
+
+Number
+
+    EGL Extension #141
+
+Extension Type
+
+    EGL client extension
+
+Dependencies
+
+    Requires EGL_EXT_client_extensions to query its existence without
+    a display.
+
+    Requires EGL_EXT_platform_base.
+
+    This extension is written against the wording of version 9 of the
+    EGL_EXT_platform_base specification.
+
+Overview
+
+    This extension defines how to create EGL resources from native X11
+    resources using the functions defined by EGL_EXT_platform_base.
+
+    The native X11 resources required by this extension are xcb resources.
+    All X11 types discussed here are defined by the header `xcb.h`.
+
+New Types
+
+    None
+
+New Procedures and Functions
+
+    None
+
+New Tokens
+
+    Accepted as the <platform> argument of eglGetPlatformDisplayEXT:
+
+        EGL_PLATFORM_XCB_EXT                    0x31DC
+
+    Accepted as an attribute name in the <attrib_list> argument of
+    eglGetPlatformDisplayEXT:
+
+        EGL_PLATFORM_XCB_SCREEN_EXT             0x31DE
+
+Additions to the EGL Specification
+
+    None.
+
+New Behavior
+
+    To determine if the EGL implementation supports this extension, clients
+    should query the EGL_EXTENSIONS string of EGL_NO_DISPLAY.
+
+    This extension defines the same set of behaviors as the X11 platform
+    extension, except Xlib types are replaced with xcb types.
+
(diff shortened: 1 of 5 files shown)


stonesthrow:
@yshui Thank you for your PR. I will try to get some xcb reviewers for feedback.

fooishbar:
Text looks good to me.

stonesthrow:
@oddhack Please check this for proper reservation of tokens.

stonesthrow:
This looks good to me. It would be good to get a few more thumbs up from xcb users to represent EXT mutivendor extension.

fooishbar:
cc @cubanismo

caramelli:
Thank you for the proposal of this extension, I have been using it for many years (you can take a look at the yagears project https://github.com/caramelli/yagears which illustrates the use of EGL_EXT_platform_xcb). But I never took the time to submit it ...
In addition, it gives consistency with the Vulkan WSI layer which supports both Xlib (VK_KHR_xlib_surface) and XCB (VK_KHR_xcb_surface)

Nicolas Caramelli

yshui:
@caramelli 

> Just note that eglplatform.h needs to be updated with something like this:

No, `xcb_connection_t *` is not going to be accepted by `eglGetDisplay`

caramelli:
I don't understand why: eglGetDisplay accepts a NativeDisplayType
eglplatform.h should normally provide a way to define EGLNativeDisplayType, EGLNativePixmapType, EGLNativeWindowType for the XCB platform.

yshui:
@caramelli To not complicate things.

`eglGetDisplay` doesn't know what is passed, so it has to use less than elegant heuristics to guess the type of the NativeDisplay. Adding a new type will make it more complicated.

caramelli:
Yes, using EXT_platform_base with eglGetPlatformDisplayEXT () does not require knowing the native display type (compared to eglGetDisplay) because it is a "void *": EGL_EXT_platform_xcb will be used like that!
But when using a platform that only has XCB (for example on an embedded system), the build fails with this error: "Platform not recognized" (unless you set the flag EGL_NO_X11 ...)
So I think it makes sense to have a way to define EGLNativeDisplayType, EGLNativePixmapType, EGLNativeWindowType for XCB, it's also consistent with other platforms and EGL history.
But this is only my feedback and to share my point of view.
Anyway, thank you again for this extension proposal!

Nicolas Caramelli

yshui:
Since the feedback are positive, I removed the draft status.

stonesthrow:
If we can get a couple reviews to represent the xcb community for EXT then it seems it would be good for merge.

extensions/EXT/EGL_EXT_platform_xcb.txt
```diff
@@ -0,0 +1,357 @@
...
+    <native_display> parameter specifies the X11 display connection to use, and
+    must point to a valid xcb `xcb_connection_t` or be NULL.  If <native_display>
+    is EGL_DEFAULT_DISPLAY, then EGL will create [1] a connection to the default
+    X11 display. The environment variable DISPLAY determines the default X11
+    display as described in the documentation of `xcb_connect`.  The value of
+    attribute EGL_PLATFORM_XCB_SCREEN_EXT specifies the X11 screen to use. If
+    the attribute is omitted from <attrib_list>, then the environment varialbe
+    DISPLAY will be used to select a screen.  Otherwise, the attribute's value
+    must be a valid screen on the display connection. If the attribute's value
+    is not a valid screen, then an EGL_BAD_ATTRIBUTE error is generated.
```
fooishbar, line 93: The X11 protocol has the concept of a 'default screen', expressed through XCB as well. We should specify the following behaviour:
* if a `XCB_SCREEN` is specified, then this screen number will always be used; else
* if a `native_display` is specified, then the default screen for that display will be used; else
* if `DISPLAY` is present and contains a screen part (e.g. `:0.1` specifies a screen, `:0` does not), then that screen part will be used; else
* the default screen for that display will be used

fooishbar:
I'm more ex-11 than X11, but at least from a Mesa winsys point of view I'm perfectly happy with this extension after the semantics around Screen specification. Thanks @yshui!

extensions/EXT/EGL_EXT_platform_xcb.txt
psychon: I'd say that the X11 protocol itself has no concept of a default screen. Only the code for parsing the `$DISPLAY` environment variable does. Also, the only way to get this default screen from xcb is via `xcb_connect`. Given an `xcb_connection_t*`, you cannot figure out the `.42` in `DISPLAY=:1.42` since it is not saved anywhere (and the env var might have changed since the connection was constructed).

Or am I missing something?

fooishbar: Right you are, I apparently can't read Xlib code anymore. Which is probably a good thing. Regardless, `xcb_parse_display` forces screen 0 if no screen is explicitly specified, and I'd suggest encoding the same here, if you supply a pointer to an `xcb_connection_t` but no screen number.

psychon: Which makes me wonder... what are the semantics of this screen number? Is it assumed that all windows are on that screen (= are a descendant of the screen's root window)? Or why does a GL implementation need to know a screen number?

cubanismo: Different screens may be driven by different GPUs/different driver vendors.  It's needed for dispatching at the very least.

XCB's lack of a default screen (It has one, it just isn't available besides xcb_connect) is quite annoying IMHO.

extensions/EXT/EGL_EXT_platform_xcb.txt
```diff
@@ -0,0 +1,357 @@
...
+
+    None.
+
+New Behavior
+
+    To determine if the EGL implementation supports this extension, clients
+    should query the EGL_EXTENSIONS string of EGL_NO_DISPLAY.
+
+    This extension defines the same set of behaviors as the X11 platform
+    extension, except Xlib types are replaced with xcb types.
```
cubanismo, line 80: I would replace "the X11 platform extension" with "EGL_EXT_platform_x11."

extensions/EXT/EGL_EXT_platform_xcb.txt
```diff
@@ -0,0 +1,357 @@
...
+
+    To obtain an EGLDisplay backed by an X11 screen, call
+    eglGetPlatformDisplayEXT with <platform> set to EGL_PLATFORM_XCB_EXT. The
+    <native_display> parameter specifies the X11 display connection to use, and
+    must point to a valid xcb `xcb_connection_t` or be NULL.  If <native_display>
```
cubanismo, line 85: NULL, or EGL_DEFAULT_DISPLAY?

extensions/EXT/EGL_EXT_platform_xcb.txt
```diff
@@ -0,0 +1,357 @@
...
+    is EGL_DEFAULT_DISPLAY, then EGL will create [1] a connection to the default
+    X11 display. The environment variable DISPLAY determines the default X11
+    display as described in the documentation of `xcb_connect`.  The value of
+    attribute EGL_PLATFORM_XCB_SCREEN_EXT specifies the X11 screen to use. If
+    the attribute is omitted from <attrib_list>, then the environment varialbe
+    DISPLAY will be used to select a screen.  Otherwise, the attribute's value
+    must be a valid screen on the display connection. If the attribute's value
+    is not a valid screen, then an EGL_BAD_ATTRIBUTE error is generated.
+
+    [fn1] The method by which EGL creates a connection to the default X11
```
cubanismo, line 95: There may be a subtle interaction with EGL_NV_native_query here: Whatever method EGL uses, it must be capable of generating an xcb_connection_t to return to the app form it if EGL_NV_native_query is supported.

extensions/EXT/EGL_EXT_platform_xcb.txt
cubanismo: Going back to @fooishbar's proposal, is the best we can do with XCB then this:

* if a XCB_SCREEN is specified, then this screen number will always be used; else
* if a native_display is specified, then screen 0 will be used.
* if native_display is EGL_DEFAULT_DISPLAY, and DISPLAY is present and contains a screen part (e.g. :0.1 specifies a screen, :0 does not), then that screen part will be used; else
* screen is undefined (But AFAIK, the display creation operation will fail anyway unless EGL has a magic way to find an X server, such as assuming DISPLAY=:0.0).

extensions/EXT/EGL_EXT_platform_xcb.txt
```diff
@@ -0,0 +1,357 @@
...
+    eglCreatePlatformWindowSurfaceEXT with a <dpy> that belongs to X11 and
+    a <native_window> that points to an xcb_window_t.
+
+    To obtain an offscreen rendering surface from an X11 Pixmap, call
+    eglCreatePlatformPixmapSurfaceEXT with a <dpy> that belongs to X11 and
+    a <native_pixmap> that points to an xcb_pixmap_t.
+
+Issues
+
+    None so far.
```
cubanismo, line 109: Probably worth listing the default screen issue here.

cubanismo (changes requested):

extensions/EXT/EGL_EXT_platform_xcb.txt
yshui: This is the same language used in `EXT_platform_x11`.
Though I think `EGL_DEFAULT_DISPLAY` is the better one here.

extensions/EXT/EGL_EXT_platform_xcb.txt
yshui: Makes sense to me.

extensions/EXT/EGL_EXT_platform_xcb.txt
cubanismo: Ah, yes, it would probably be best to update both, but at least we can get it right here to start.

extensions/EXT/EGL_EXT_platform_xcb.txt
yshui: Maybe it's better to explicitly say the display creation operation would fail in the last case?

yshui:
Screen selection issue is addressed.

I still need to look into EGL_NV_native_query...

extensions/EXT/EGL_EXT_platform_xcb.txt
yshui: This is the same language used in `EXT_platform_x11` too.

~~Mesa stores a `xcb_connection_t *` so this won't be a problem for them. If other drivers don't do this, they probably have to choose between not supporting `EXT_platform_xcb` or not supporting `NV_native_query`.~~

Oh wait, this extension doesn't define a `EGLNativeDisplayType`, which means `NV_native_query` have to create a Xlib connection from a `xcb_connection_t`.

I think it's probably best to update `NV_native_query` to say, the native display can only be retrieved if a `EGLNativeDisplayType` is used to create the `EGLDisplay`.

cubanismo: I'm not clear why that is.  The native display type is a property of the EGLDisplay.  NV_native_query implicitly returns an object of the type corresponding to the display's native display type, right?

I agree this is likely an issue in EXT_platform_x11 as well, but I think simply noting an interaction as you originally defined above (If you do something crazy, you can't support both extensions) in the "dependencies" section would be fine to resolve this, and then we  could backport that to platform_x11.

extensions/EXT/EGL_EXT_platform_xcb.txt
cubanismo: I don't know.  EGL implementations have historically taken a lot of liberties here.  It seems worth leaving it undefined in case some weird use case requires the EGL library to do something clever here would be harmless.  Apps != conformance tests rarely complain about errors not showing up exactly as expected in crazy corner cases in my experience.

extensions/EXT/EGL_EXT_platform_xcb.txt
yshui: NV_native_query explicitly specifies that the function would return a `EGLNativeDisplayType`, which is the type passed to `eglGetDisplay`. This extension doesn't define a `EGLNativeDisplayType`.

extensions/EXT/EGL_EXT_platform_xcb.txt
yshui: OK, I will update.

extensions/EXT/EGL_EXT_platform_xcb.txt
cubanismo: Yes, this is a rough edge, but I think in practice EGLNativeDisplayType will always be a pointer, and is effectively void* in any implementation that supports multiple platforms, so very pedantic C implementations (are there any C implementations where pointer types actually differ besides between function pointers and data pointers?) aside, I would assume it's just a matter of casting EGLNativeDisplayType back to the appropriate type for your platform.

yshui: I can take this interpretation. So I guess this is a non-issue then?

cubanismo: Pretty much. I think it's worth mentioning as an interaction issue (using text similar to suggested above), but if you disagree, not a big deal. It's a pretty dark corner case.

cubanismo (approved):
Approved from my POV, with or without a mention of the NV_native_query interaction.

stonesthrow (approved):

PR #114, opened 2020-10-01
api: Fix genheaders.py to emit valid comments

nwnk:
When we switched to using SPDX identifiers instead of explicit Apache
2.0 text, the resulting generated headers ended up with a nested /*
inside the license comment.


stonesthrow:
@oddhack - to review changes in markup.

oddhack:
@jackohound LGTM. Mark Approved when ready.

stonesthrow (approved):

PR #121, opened 2021-01-26
Add EGL_MESA_device_query_type

emersion:
This is inspired by [`VkPhysicalDeviceType`][1] and lets clients
query the type of an EGL device.

[1]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDeviceType.html

* * *

I'd like to know whether this is a good idea before going forward with the reservations and the spec text.

I am not a Khronos member. I have no plans to make this an EXT or KHR extension.

cc @cubanismo @kbrenneman @nwnk

### DIFF
--- a/api/egl.xml
+++ b/api/egl.xml
@@ -1025,6 +1025,16 @@
         <unused start="0x34B0" end="0x34BF"/>
     </enums>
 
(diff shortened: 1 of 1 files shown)


emersion:
It turns out `EGL_MESA_device_software` exists, so I don't need this anymore. Sorry for the noise.

PR #122, opened 2021-02-02
extensions: Add EGL_EXT_device_persistent_id

kbrenneman:
This is a new extension which provides a persistent identifier string for an `EGLDeviceEXT`.

The primary purpose is so that if an application selects a device, it can record this device ID string in a config file. When it runs again later, it can use that ID to find the same device again.

In addition, the device ID string contains a vendor name as a prefix. Mostly, that just ensures uniqueness on systems that allow multiple ICD's, but it would also allow you to correlate that driver to other API's. In particular, XWayland could use that to pick a GLX_VENDOR_NAMES_EXT string to send to clients based on the EGLDisplay that it's using to talk to the compositor.

### DIFF
--- a/api/EGL/eglext.h
+++ b/api/EGL/eglext.h
@@ -14,12 +14,12 @@ extern "C" {
 ** used to make the header, and the header can be found at
 **   http://www.khronos.org/registry/egl
 **
-** Khronos $Git commit SHA1: 5a9a7e3fcb $ on $Git commit date: 2020-08-24 11:05:32 -0700 $
+** Khronos $Git commit SHA1: 59522adade $ on $Git commit date: 2021-02-02 11:09:11 -0700 $
 */
 
 #include <EGL/eglplatform.h>
 
-#define EGL_EGLEXT_VERSION 20201001
+#define EGL_EGLEXT_VERSION 20210419
 
 /* Generated C header for:
  * API: egl
@@ -578,7 +578,7 @@ EGLAPI EGLBoolean EGLAPIENTRY eglQuerySurfacePointerANGLE (EGLDisplay dpy, EGLSu
 #define EGL_ANGLE_sync_control_rate 1
 typedef EGLBoolean (EGLAPIENTRYP PFNEGLGETMSCRATEANGLEPROC) (EGLDisplay dpy, EGLSurface surface, EGLint *numerator, EGLint *denominator);
 #ifdef EGL_EGLEXT_PROTOTYPES
-EGLAPI EGLBoolean EGLAPIENTRY eglGetMscRateANGLE(EGLDisplay dpy, EGLSurface surface, EGLint *numerator, EGLint *denominator);
+EGLAPI EGLBoolean EGLAPIENTRY eglGetMscRateANGLE (EGLDisplay dpy, EGLSurface surface, EGLint *numerator, EGLint *denominator);
 #endif
 #endif /* EGL_ANGLE_sync_control_rate */
 
@@ -691,6 +691,13 @@ EGLAPI EGLBoolean EGLAPIENTRY eglQueryDisplayAttribEXT (EGLDisplay dpy, EGLint a
 #define EGL_EXT_device_openwf 1
 #define EGL_OPENWF_DEVICE_ID_EXT          0x3237
 #endif /* EGL_EXT_device_openwf */
+
+#ifndef EGL_EXT_device_persistent_id
+#define EGL_EXT_device_persistent_id 1
+#define EGL_DEVICE_UUID_EXT               0x335C
+#define EGL_DRIVER_UUID_EXT               0x335D
+#define EGL_DRIVER_NAME_EXT               0x335E
+#endif /* EGL_EXT_device_persistent_id */
 
 #ifndef EGL_EXT_device_query
 #define EGL_EXT_device_query 1
--- a/extensions/EXT/EGL_EXT_device_persistent_id.txt
+++ b/extensions/EXT/EGL_EXT_device_persistent_id.txt
@@ -89,6 +89,10 @@ Changes to section 3.2 (Devices)
     <max_size> is ignored, and the size of the attribute in bytes is
     returned in <size>.
 
+    On failure, EGL_FALSE is returned. An EGL_BAD_ATTRIBUTE error is
+    generated if <name> is not a valid attribute. An EGL_BAD_DEVICE_EXT
+    error is generated if <device> is not a valid EGLDeviceEXT.
+
     If <value> is not NULL, then the attribute value is returned in
     <value>. At most <max_size> bytes are written. <size> returns the
     number of bytes that were actually written.
--- a/extensions/EXT/EGL_EXT_device_persistent_id.txt
+++ b/extensions/EXT/EGL_EXT_device_persistent_id.txt
@@ -1,0 +1,198 @@
+Name
+
+    EXT_device_persistent_id
+
+Name Strings
+
+    EGL_EXT_device_persistent_id
+
+Contributors
+
+    Kyle Brenneman,  NVIDIA  (kbrenneman 'at' nvidia.com)
+
+Contact
+
+    Kyle Brenneman,  NVIDIA  (kbrenneman 'at' nvidia.com)
+
+Version
+
+    Version 1 - April 19, 2021
+
+Number
+
+    EGL Extension #142
+
+Extension Type
+
+    EGL device extension
+
+Dependencies
+
+    Written based on the wording of the EGL 1.5 specification.
+
+    EGL_EXT_device_query is required.
+
+Overview
+
+    Applications can query a list of EGLDeviceEXT handles, but those
+    handles are only valid within the process that queried them. An
+    application has no way, for example, to record its selection and
+    select the same device when run again later.
+
+    This extension provides a vendor name and a set of UUID's, which
+    provide a unique, persistent identifier for EGLDeviceEXT handles.
+    This allows applications to find the EGLDeviceEXT for the same
+    device across multiple processes, and across multiple APIs.
+
+New Procedures and Functions
+
+    EGLBoolean eglQueryDeviceBinaryEXT(EGLDeviceEXT device,
+                                      EGLint name,
+                                      EGLint max_size,
+                                      void *value,
+                                      EGLint *size);
+
+New Tokens
+
+    Accepted by the <name> parameter of eglQueryDeviceStringEXT:
+
+        EGL_DRIVER_NAME_EXT          0x335E
+
+    Accepted by the <name> parameter of eglQueryDeviceBinaryEXT:
+
+        EGL_DEVICE_UUID_EXT          0x335C
+        EGL_DRIVER_UUID_EXT          0x335D
+
+Changes to section 3.2 (Devices)
+
+    Add the following paragraph to the description of
+    eglQueryDeviceStringEXT:
+
+    EGL_DRIVER_NAME_EXT returns a string which identifies the driver
+    that controls the device. This string remains persistent across
+    multiple versions of a driver, and an application can use strcmp(3)
+    to compare the strings for equality. Otherwise, the contents are
+    implementation-defined.
+
+
+    Add to the end of section 3.2:
+
+    To query a binary attribute for a device, use:
+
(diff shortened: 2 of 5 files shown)


api/EGL/eglext.h
```diff
@@ -578,7 +578,7 @@ EGLAPI EGLBoolean EGLAPIENTRY eglQuerySurfacePointerANGLE (EGLDisplay dpy, EGLSu
 #define EGL_ANGLE_sync_control_rate 1
 typedef EGLBoolean (EGLAPIENTRYP PFNEGLGETMSCRATEANGLEPROC) (EGLDisplay dpy, EGLSurface surface, EGLint *numerator, EGLint *denominator);
 #ifdef EGL_EGLEXT_PROTOTYPES
-EGLAPI EGLBoolean EGLAPIENTRY eglGetMscRateANGLE(EGLDisplay dpy, EGLSurface surface, EGLint *numerator, EGLint *denominator);
+EGLAPI EGLBoolean EGLAPIENTRY eglGetMscRateANGLE (EGLDisplay dpy, EGLSurface surface, EGLint *numerator, EGLint *denominator);
```
stonesthrow, line 581: Why is space added?

stonesthrow (commented):
Looks like a useful extension. Get a few others ISV to justify EXT.

api/EGL/eglext.h
kbrenneman: That's just what came out of the script when it regenerated the header.

The space is consistent with the other functions, so it's more odd that it didn't have a space before.

If you want, I could split off a separate commit at the beginning which just regenerates eglext.h from an unmodified egl.xml. That would add the missing space, and keep things a little better separated.

emersion:
How does this relate to [GL](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_external_objects.txt) and [Vulkan](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPhysicalDeviceIDProperties.html) device UUIDs? The Vulkan UUID is already spec'ed to be immutable across reboots.

> deviceUUID must be immutable for a given device across instances, processes, driver APIs, driver versions, and system reboots.

It would be nice if this allowed to build a map between EGLDevice and VkPhysicalDevice. Some clients may want to use both GL and Vulkan, making sure the same physical device is used with both APIs is important in this case.

kbrenneman:
The device ID string is vendor-defined and opaque, so a vendor can use a Vulkan UUID, but it's not required to. Nor is this limited to devices that support Vulkan, for that matter. I wrote it up that way to minimize the assumptions about how a vendor library might want to define an identifier.

Aside from that, the deviceUUID string in Vulkan is defined more for correlating devices between API's, not for being able to record a device selection in a config file and looking it up later. From the Vulkan spec:

> While implementations should make every effort to report consistent device UUID's across driver versions, applications should avoid relying on the persistence of this value for uses other than identifying compatible devices for external object sharing purposes.

Also, the device ID string is encouraged, but not required to be persistent across driver versions. It's possible that fixing some driver bug might necessitate changing how the driver generates device ID's.

Looking up the Vulkan driver/device UUID's for an EGLDeviceEXT could be a worthwhile extension in its own right, though.

emersion:
> limited to devices that support Vulkan

I don't think these UUIDs force the driver to have a Vulkan implementation.

> I wrote it up that way to minimize the assumptions about how a vendor library might want to define an identifier.

I was hoping we could avoid avoid ending up with yet another brand new identifier...

What is the motivation for this? What are the drivers that wouldn't be able to generate a proper UUID?

> Aside from that, the deviceUUID string in Vulkan is defined more for correlating devices between API's, not for being able to record a device selection in a config file and looking it up later.

Your spec text (e.g. https://github.com/KhronosGroup/EGL-Registry/pull/122/files#diff-35f33a5ef95e3afe39640737128c2aafbcdcacae375ef9c90a7e095fdcea84f1R62) doesn't seem to provide more guarantees than Vulkan UUIDs. In fact, it seems to provide the exact same guarantees.

How do you address the concerns explained in the note in the Vulkan spec?

> Also, the device ID string is encouraged, but not required to be persistent across driver versions. 

Spec doesn't seem to indicate this: "deviceUUID must be immutable for a given device across […] driver versions".

> In particular, XWayland could use that to pick a GLX_VENDOR_NAMES_EXT string to send to clients based on the EGLDisplay that it's using to talk to the compositor.

Sounds like this is about "identifying compatible devices", not storage in configuration files. There's no guarantee that the compositor is using EGL. For this use-case, a device+driver UUID is better.

kbrenneman:
My reason for using an opaque string instead of Vulkan's UUID's was to make the interface simpler for applications, and to make as few assumptions as possible about a driver's implementation of the extension.

Querying and storing a single opaque string doesn't require any new EGL functions, and it's much simpler and easier for an application to store a single string then having to store a tuple of values. Internally, an opaque string gives a driver the best flexibility about generating that string.

Now that I've since read a little more about UUID formats, though, a UUID actually doesn't make more assumptions than a string would. If you can generate a string, then you can hash that string to generate a version 3 or 5 UUID. So, from the driver's perspective, a UUID is probably fine.

For the purpose of letting an application record a device selection, a UUID would be fine as long as the device UUID *by itself* is sufficient to uniquely identify an EGLDeviceEXT, and so an application only needs to store a single value, not a tuple. That means that two drivers would need to produce different device UUID's even if they're talking to the same physical device, which (I think?) is a stronger uniqueness requirement than Vulkan has.

Also, translating an EGLDeviceEXT to a GLX vendor library for the GLX_VENDOR_NAMES_EXT string would require a vendor name, *not* a UUID. The GLX vendor name is used to construct a filename to pass to dlopen. GLX doesn't have any notion of devices, so after libglvnd loads a vendor library, it's really up to that vendor library to pick a device internally.

So, based on the above, if we wanted to use UUID's so that an application could both record a device selection and translate between EGLDeviceEXT handles, VkPhysicalDevice handles, and GLX vendor libraries, then I think we could. We'd need to add a new EGL function (to query binary device attributes), and three new query enums (device UUID, driver UUID, and vendor name).

kbrenneman:
Looking more at Vulkan's UUID's, using them as currently defined would not be sufficient for this extension.

The goal of this extension is to provide a persistent, unique identifier for each EGLDeviceEXT.

If I'm reading the Vulkan docs correctly, then `deviceUUID` is persistent, but *not* unique. If you've got the same physical device with two different drivers, then they could both use the same `deviceUUID`, but with different `driverUUID` values. Hence, why you always have to compare both the `deviceUUID` and `driverUUID`, never only one or the other.

The `driverUUID` is explicitly defined to not be persistent. It's supposed to be different with every driver version. Therefore, a `(driverUUID, deviceUUID)` pair is also not persistent.

So, if we want to be able to use this extension to correlate to Vulkan UUID's, then we'd still need to define a new persistent driver identifier. Using a string as a vendor name would work for that, and would also then cleanly correlate to GLX.

emersion:
> If I'm reading the Vulkan docs correctly, then deviceUUID is persistent, but not unique.

Hmm, my reading would suggest that a `deviceUUID` _should_ be unique:

> Implementations should return deviceUUID values which are likely to be unique even in the presence of multiple Vulkan implementations. […] A combination of values unique to the vendor, the driver, and the hardware environment can be used to provide a deviceUUID which is unique to a high degree of certainty.

Although the Vulkan spec is a little evasive because it's impossible to guarantee uniqueness when drivers can choose any value they want.

> The driverUUID is explicitly defined to not be persistent. It's supposed to be different with every driver version.

Are you referring to "universally unique identifier _for the driver build_"?

> So, if we want to be able to use this extension to correlate to Vulkan UUID's, then we'd still need to define a new persistent driver identifier. Using a string as a vendor name would work for that, and would also then cleanly correlate to GLX.

Is `GL_RENDERER` not suitable for this?

kbrenneman:
> > If I'm reading the Vulkan docs correctly, then deviceUUID is persistent, but not unique.
> 
> Hmm, my reading would suggest that a `deviceUUID` _should_ be unique:
> 
> > Implementations should return deviceUUID values which are likely to be unique even in the presence of multiple Vulkan implementations. […] A combination of values unique to the vendor, the driver, and the hardware environment can be used to provide a deviceUUID which is unique to a high degree of certainty.
> 
> Although the Vulkan spec is a little evasive because it's impossible to guarantee uniqueness when drivers can choose any value they want.

It's ambiguous to me whether that uniqueness requirement applies to the VkPhysicalDevice handle, or only to physical device itself. That is: If two drivers can support the same physical device, then the application would get two different VkPhysicalDevice handles, one from each driver. Each VkPhysicalDevice would have a different driverUUID value (since they come from different drivers), but are they allowed to use the same deviceUUID?

> 
> > The driverUUID is explicitly defined to not be persistent. It's supposed to be different with every driver version.
> 
> Are you referring to "universally unique identifier _for the driver build_"?

Correct. Part of the goal for this extension is for an application to store a device identifier in a config file, and then be able to find the EGLDeviceEXT handle for that same device again later. Since the driverUUID value changes every time you update a driver, that means an identifier that includes driverUUID would become invalid.

> 
> > So, if we want to be able to use this extension to correlate to Vulkan UUID's, then we'd still need to define a new persistent driver identifier. Using a string as a vendor name would work for that, and would also then cleanly correlate to GLX.
> 
> Is `GL_RENDERER` not suitable for this?

No, it's not. For GLX, what I'm talking about is this extension:
https://www.khronos.org/registry/OpenGL/extensions/EXT/GLX_EXT_libglvnd.txt

The GLX_VENDOR_NAMES_EXT string specifies a driver using a fixed string, like `"nvidia"` or `"mesa"`. Libglvnd's GLX client library uses that string to know which vendor library it should load.

For this extension, if two drivers using the same physical device can use the same deviceUUID, then using that same vendor name would let an application uniquely identify an EGLDeviceEXT handle.

emersion:
OK, thanks for the explanation. I don't have anything else to add at this time.

kbrenneman:
Sorry for the delay on this. I've pushed a new version of the spec that uses UUID's to identify devices.

The new version defines three new queries, plus a new function.

The `EGL_DEVICE_UUID_EXT` and `EGL_DRIVER_UUID_EXT` attributes correspond to the `deviceUUID` and `driverUUID` fields in Vulkan's `VkPhysicalDeviceIDProperties` struct. The extension adds a new `eglQueryDeviceBinaryEXT` function to query them.

There's also a new `EGL_DRIVER_NAME_EXT` attribute that you can query with `eglQueryDeviceStringEXT`. An application can use a `(EGL_DRIVER_NAME_EXT, EGL_DEVICE_UUID_EXT)` pair as a persistent, unique identifier. `EGL_DEVICE_UUID_EXT` alone isn't unique (two drivers that support the same GPU could use the same device UUID), and `EGL_DRIVER_UUID_EXT` is by definition not persistent.

On systems that use libglvnd, `EGL_DRIVER_NAME_EXT` would also correspond to a vendor name in `GLX_VENDOR_NAMES_EXT`.

extensions/EXT/EGL_EXT_device_persistent_id.txt
```diff
@@ -0,0 +1,198 @@
...
+
+        EGLBoolean eglQueryDeviceBinaryEXT(EGLDeviceEXT device,
+                                          EGLint name,
+                                          EGLint max_size,
+                                          void *value,
+                                          EGLint *size);
+
+    On success, EGL_TRUE is returned. If <value> is NULL, then
+    <max_size> is ignored, and the size of the attribute in bytes is
+    returned in <size>.
```
emersion, line 90: Can we add a small paragraph about failures? `EGL_FALSE` must be returned, `EGL_BAD_DEVICE_EXT` is generated if the device handle is invalid, `EGL_BAD_PARAMETER` is generated if the name is invalid (see `EGL_EXT_device_query`).

emersion:
Overall the spec LGTM.

extensions/EXT/EGL_EXT_device_persistent_id.txt
kbrenneman: Good point, thanks. I've added that paragraph to the spec.

stonesthrow:
Ready for merge Kyle?

kbrenneman:
I'll change the issues section from "proposed" to "resolved," and then I think it's ready to merge.

stonesthrow:
@oddhack , This is ready to merge.

PR #123, opened 2021-03-12
Fix an enum value in the EGL_EXT_device_query_name spec

kbrenneman:
This fixes a typo in the EGL_EXT_device_query_name spec. The EGL_RENDERER_EXT enum is supposed to have the value 0x335F, but I typed in 0x335E when writing up the spec.

egl.xml (and eglext.h) have the correct value.


stonesthrow:
OK, we'll have Jon verify the token and merge your fix.

stonesthrow (approved):
Thanks for the fix.

PR #124, opened 2021-04-06
extensions: Add EGL_EXT_config_select_group

rmader:
This extension will be used to allow picking EGL configs that
are handled as transparent by X11 compositors.

See also:
 - https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/9989
 - https://bugzilla.mozilla.org/show_bug.cgi?id=1702546

### DIFF
--- a/api/EGL/eglext.h
+++ b/api/EGL/eglext.h
@@ -651,6 +651,11 @@ EGLAPI EGLBoolean EGLAPIENTRY eglCompositorSwapPolicyEXT (EGLint external_win_id
 #endif
 #endif /* EGL_EXT_compositor */
 
+#ifndef EGL_EXT_config_select_group
+#define EGL_EXT_config_select_group 1
+#define EGL_CONFIG_SELECT_GROUP_EXT       0x34C0
+#endif /* EGL_EXT_config_select_group */
+
 #ifndef EGL_EXT_create_context_robustness
 #define EGL_EXT_create_context_robustness 1
 #define EGL_CONTEXT_OPENGL_ROBUST_ACCESS_EXT 0x30BF
--- a/api/egl.xml
+++ b/api/egl.xml
@@ -1030,7 +1030,8 @@
     </enums>
 
     <enums namespace="EGL" start="0x34C0" end="0x34CF" vendor="EXT" comment="Reserved for Robert Mader (PR 124)">
-        <unused start="0x34C0" end="0x34CF"/>
+        <enum value="0x34C0" name="EGL_CONFIG_SELECT_GROUP_EXT"/>
+            <unused start="0x34C1" end="0x34CF"/>
     </enums>
 
 <!-- Please remember that new enumerant allocations must be obtained by
@@ -2395,6 +2396,11 @@
                 <enum name="EGL_SYNC_CLIENT_SIGNAL_EXT"/>
                 <command name="eglClientSignalSyncEXT"/>
             </require>
+        </extension>
+        <extension name="EGL_EXT_config_select_group" supported="egl">
+          <require>
+                <enum name="EGL_CONFIG_SELECT_GROUP_EXT"/>
+          </require>
         </extension>
         <extension name="EGL_EXT_create_context_robustness" supported="egl">
             <require>
--- a/extensions/EXT/EGL_EXT_config_select_group.txt
+++ b/extensions/EXT/EGL_EXT_config_select_group.txt
@@ -1,0 +1,111 @@
+Name
+
+    EGL_EXT_config_select_group
+
+Name Strings
+
+    EGL_EXT_config_select_group
+
+Contributors
+
+    Hal Gentz <zegentzy@protonmail.com>
+    Adam Jackson <ajax@redhat.com>
+    Robert Mader <robert.mader@posteo.de>
+
+Contacts
+
+    Hal Gentz <zegentzy@protonmail.com>
+
+Status
+
+    Draft
+
+Version
+
+    Version 4, 2021-06-24
+
+Number
+
+    EGL Extension #145
+
+Extension Type
+
+    EGL client extension
+
+Dependencies
+
+    This extension is written against the wording of the 2014.08.27 revision
+    of the EGL 1.5 Specification.
+
+Overview
+
+    This extension provides a mechanism to lower the config selection priority
+    of the configs returned by eglChooseConfig. A new config attribute is
+    introduced, providing a way for the implementation to group configs into
+    different config selection categories.  The config selection priorities of
+    all the configs in one category may be higher or lower than the ones in
+    other categories.  The config selection priorities in one category follow
+    the rules of eglChooseConfig. See also GLX_SGIX_visual_select_group.
+
+New Types
+
+    None
+
+New Procedures and Functions
+
+    None
+
+New Tokens
+
+    Accepted as a new EGLConfig attribute:
+
+        EGL_CONFIG_SELECT_GROUP_EXT                0x34C0
+
+Additions to the EGL 1.5 Specification
+
+    In section 3.4 "Configuration Management", under the "Buffer Descriptions
+    and Attributes" heading, append to table 3.1 the following:
+
+    "EGL_CONFIG_SELECT_GROUP_EXT | integer | the config select group"
+
+    In section 3.4 "Configuration Management", under the "Other EGLConfig
+    Attribute Descriptions" heading, append to the end the following:
+
+    "The attribute EGL_CONFIG_SELECT_GROUP_EXT is used to specify the config
+    selection category of the config. Configs in the i-th category have a higher
+    config selection priority than those found in the (i+1)-th category.
+
+    "The attribute EGL_CONFIG_SELECT_GROUP_EXT is used by the implementation to
+    override the default sorting rules, by collecting configurations into
+    groups. It is not accepted by the <attrib_list> parameter of
+    eglChooseConfig, but it is accepted as the <attribute> in
+    eglGetConfigAtrrib."
+
+    In section 3.4 "Configuration Management", under the "Sorting of EGLConfigs"
+    heading, append to table 3.4 the following:
+
+    "EGL_CONFIG_SELECT_GROUP_EXT | 0 | Ignore | Smaller | 0"
+
+    In section 3.4 "Configuration Management", under the "Sorting of EGLConfigs"
+    heading, append before the list entry which starts with "1. Special: by
+    EGL_CONFIG_CAVEAT" the following entry:
+
+    "0. Smaller EGL_CONFIG_SELECT_GROUP_EXT."
+
+Issues
+
+    None
+
+Revision History
+
+    Version 4, 2021-06-24 (Robert Mader)
+        - Moved to EXT, changed enum value to 0x34C0.
+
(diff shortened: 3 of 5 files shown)


api/EGL/eglext.h
```diff
@@ -578,7 +578,7 @@ EGLAPI EGLBoolean EGLAPIENTRY eglQuerySurfacePointerANGLE (EGLDisplay dpy, EGLSu
 #define EGL_ANGLE_sync_control_rate 1
 typedef EGLBoolean (EGLAPIENTRYP PFNEGLGETMSCRATEANGLEPROC) (EGLDisplay dpy, EGLSurface surface, EGLint *numerator, EGLint *denominator);
 #ifdef EGL_EGLEXT_PROTOTYPES
-EGLAPI EGLBoolean EGLAPIENTRY eglGetMscRateANGLE(EGLDisplay dpy, EGLSurface surface, EGLint *numerator, EGLint *denominator);
+EGLAPI EGLBoolean EGLAPIENTRY eglGetMscRateANGLE (EGLDisplay dpy, EGLSurface surface, EGLint *numerator, EGLint *denominator);
```
stonesthrow, line 581: White space introduced here.

extensions/MESA/EGL_MESA_config_select_group.txt
```diff
@@ -0,0 +1,108 @@
...
+    In section 3.4 "Configuration Management", under the "Other EGLConfig
+    Attribute Descriptions" heading, append to the end the following:
+
+    "The attribute EGL_CONFIG_SELECT_GROUP_MESA is used to specify the config
+    selection category of the config. Configs in the i-th category have a higher
+    config selection priority than those found in the (i+1)-th category.
+
+    "The attribute EGL_CONFIG_SELECT_GROUP_MESA is used by the implementation to
+    override the default sorting rules, by collecting configurations into
+    groups. It is not accepted by the <attrib_list> parameter of
```
stonesthrow, line 80: Not accepted with eglChooseConfig, but you are providing details on selection?

extensions/MESA/EGL_MESA_config_select_group.txt
```diff
@@ -0,0 +1,108 @@
...
+    eglChooseConfig, but it is accepted as the <attribute> in
+    eglGetConfigAtrrib."
+
+    In section 3.4 "Configuration Management", under the "Sorting of EGLConfigs"
+    heading, append to table 3.4 the following:
+
+    "EGL_CONFIG_SELECT_GROUP_MESA | 0 | Ignore | Smaller | 0"
```
stonesthrow, line 87: Add "Ignore" to  section 3.4.1.1

extensions/MESA/EGL_MESA_config_select_group.txt
```diff
@@ -0,0 +1,108 @@
...
+
+    In section 3.4 "Configuration Management", under the "Sorting of EGLConfigs"
+    heading, append before the list entry which starts with "1. Special: by
+    EGL_CONFIG_CAVEAT" the following entry:
+
+    "0. Smaller EGL_CONFIG_SELECT_GROUP_MESA."
```
stonesthrow, line 93: Is sorting by group always enabled? or what is the default behavior, and when is new sort rule 0 enabled?
This would override current sorting rules and give app different results than prior to this. This should be a clear opt-in case for the app.

stonesthrow (commented):
There are a couple things about selection that are not clear.

api/EGL/eglext.h
rmader: This header is auto-generated - so any PR that relies on running `make` in `/api/` will produce this. I can create a independent PR or commit for it if you prefer.

stonesthrow: Ah, of course. I see now its throughout.  Sorry. Ignore comment.

extensions/MESA/EGL_MESA_config_select_group.txt
stonesthrow: Not to dissuade you from your design, But have you thought about co-opting the EGL_NATIVE_VISUAL_TYPE attribute for a X11/MESA special rule - exact match for anything greater than EGL_DONT_CARE. Then if attribute_list has EGL_NATIVE_VISUAL_TYPE, EGL_CONFIG_TRANSPARENT_MESA (new token). Then only those configs would be returned - same sorting order. Selection rule being "Exact" match, sorting none?. This attribute is limited to windows surfaces (section 3.4.1.1). Maybe that gets you what you want?

extensions/MESA/EGL_MESA_config_select_group.txt
nwnk: This was at my suggestion, primarily for diagnosing why you got the config you did. I'm not sure it's the best plan, the corresponding GLX extn doesn't let you get this from the app, it's just a protocol detail. I'd be just as happy to drop that bit I think.

extensions/MESA/EGL_MESA_config_select_group.txt
nwnk: `EGL_NATIVE_VISUAL_TYPE` is already used to communicate the difference between TrueColor and PseudoColor (or StaticColor or DirectColor or...) visuals in X11. The "transparent" visual under X is just one whose depth exceeds the number of r/g/b bits.

nwnk:
The background here is that under X the config select rules push you onto configs that have alpha even if you didn't ask for it, and then your windows are accidentally translucent, so we started using `GLX_SGIX_visual_select_group` to punish the alphaful visual so you wouldn't get it unless you asked for non-zero alpha bits. And then a similar problem happens under EGL, so we ported the extn across. IIRC the EGL spec has some language about implementations are encouraged but not required to sort configs in such-and-such a way, and this extn is mostly here to allow the app to know that this is Mesa and that Meaa deviates from those rules for pretty good reasons.

stonesthrow:
I see, there is more going on than I am aware of. 
I've poked the EGL workgroup to get more eyes and feedback for you.

api/egl.xml
```diff
@@ -3027,6 +3027,11 @@
                 <command name="eglWaitSyncKHR"/>
             </require>
         </extension>
+        <extension name="EGL_MESA_config_select_group" supported="egl">
+          <require>
+                <enum name="EGL_CONFIG_SELECT_GROUP_MESA"/>
+            </require>
```
cubanismo, line 3033: Indentation seems off here.

extensions/MESA/EGL_MESA_config_select_group.txt
cubanismo: The justification seems fine IMHO. I can't think of use cases outside of automated testing and printf debugging, but those are real use cases.

cubanismo (commented):
Minor formatting issue noted inline. Would be fine with making this an EXT with NV's rubber stamp as well.

extensions/MESA/EGL_MESA_config_select_group.txt
```diff
@@ -0,0 +1,108 @@
...
+
+Overview
+
+    This extension provides a mechanism to lower the config selection priority
+    of the configs returned by eglChooseConfig. A new config attribute is
+    introduced, providing a way for the implementation to group configs into
+    different config selection categories.  The config selection priorities of
+    all the configs in one category may be higher or lower than the ones in
+    other categories.  The config selection priorities in one category follow
+    the rules of eglChooseConfig.
```
ianromanick, line 48: It's probably worth explicitly referencing the GLX extension in this extension.  Possibly just add something here like

```
    See also GLX_SGIX_visual_select_group.
```

rmader:
Sorry for the long delay - now GTK4 is affected as well, so we should bring this forward.

> Would be fine with making this an EXT with NV's rubber stamp as well.

Is there anyone from Nvidia we can poke so we can make this EXT?

cubanismo:
> Is there anyone from Nvidia we can poke so we can make this EXT?

From my above comment:

> Would be fine with making this an EXT with NV's rubber stamp as well.

Up to the authors. NV is fine with this becoming an EXT.

stonesthrow:
Resolve conflicts. Address any feedback, and when ready ping me and we'll change to ready to merge and assign editor to check and merge.

extensions/MESA/EGL_MESA_config_select_group.txt
rmader: Done

rmader:
Thanks everyone, made it EXT now.

@cubanismo: I still haven't heard of the "owner" of the previously used enum value and thus took the freedom to take one from you - is that ok (and this specific one)?

@stonesthrow: ping :)

cubanismo:
> Thanks everyone, made it EXT now.

Cool.
 
> @cubanismo: I still haven't heard of the "owner" of the previously used enum value and thus took the freedom to take one from you - is that ok (and this specific one)?

Unfortunately, that value (0x3320) is already used for pending/internal extensions at NVIDIA. I'd just reserve Mesa some by editing the XML as Jeff suggests, but if you'd like, I can reserve 0x3378 from our range for this extension as a one-off.

rmader:
Ok, trying to reserve my own range in https://github.com/KhronosGroup/EGL-Registry/pull/128 now, hope there won't be any mid-air collisions.

extensions/EXT/EGL_EXT_config_select_group.txt
```diff
@@ -0,0 +1,111 @@
...
+
+    "0. Smaller EGL_CONFIG_SELECT_GROUP_EXT."
+
+Issues
+
+    None
+
+Revision History
+
+    Version 3, 2021-06-24 (Robert Mader)
```
stonesthrow, line 101: version 4

stonesthrow (changes requested):

extensions/EXT/EGL_EXT_config_select_group.txt
rmader: Of course.

stonesthrow: Nit pick. Sorry

stonesthrow (approved):

extensions/EXT/EGL_EXT_config_select_group.txt
rmader: Please don't spare me with those, better now than never.

stonesthrow:
Can we get another review approval?

rmader:
> Can we get another review approval?

@cubanismo maybe? :) Now that NV can use DMABUF, GLX to EGL transition on X11 become ever more valuable.

Darkspirit:
@rmader Do you know if Nvidia supports a sufficient vsync method on EGL/X11 (to not have 60 Hz hardcoded)?
https://gitlab.freedesktop.org/mesa/mesa/-/issues/2930

rmader:
> @rmader Do you know if Nvidia supports a sufficient vsync method on EGL/X11 (to not have 60 Hz hardcoded)?
> https://gitlab.freedesktop.org/mesa/mesa/-/issues/2930

Not that I know, no. I tend to think: if you want proper refresh rates - switch to Wayland.

cubanismo (approved):

stonesthrow:
It seems ready to go.

rmader:
Thanks everyone!

PR #128, opened 2021-06-24
Reserve enums 0x34C0..0x34CF for Robert Mader (for #124)

rmader:
and potential future extensions.

cc @oddhack, @stonesthrow


rmader:
Removed stray new line.

oddhack:
Looks fine. If this is coming from the Mesa project then we may want to switch the reservation into name so other Mesa extensions can draw from it in the future, but it's not a big deal.

PR #130, opened 2021-08-05
Don't use Xlib headers by default

kbrenneman:
This changes eglplatform.h so that it'll use the `EGL_NO_X11` typedefs by default.

This has come up a couple times in libglvnd bug reports, where applications that use EGL but not X11 don't build on systems where the Xlib headers aren't installed. The `EGL_NO_X11` macro can fix that, but adding that to every broken app out there is a tall order, and adding it to libglvnd's pkg-config files seems like a crude workaround at best.

This change should be pretty safe, I think. Applications that currently pass a `Display *` value to eglGetDisplay will still build, because you can always pass a `Display *` to a `void *`. `EGLNativePixmapType` and `EGLNativeWindowType` won't change at all, because in both the Xlib and non-Xlib cases, they're typedefs for `unsigned long`.

Existing calls to eglQueryNativeDisplayNV could break, but I'd expect those to be relatively uncommon, and would just need a trivial typecast to fix. Alternately, you can still define `USE_X11` to get the current behavior.

### DIFF
--- a/api/EGL/eglplatform.h
+++ b/api/EGL/eglplatform.h
@@ -103,13 +103,7 @@ typedef intptr_t EGLNativeDisplayType;
 typedef intptr_t EGLNativePixmapType;
 typedef intptr_t EGLNativeWindowType;
 
-#elif defined(__unix__) && defined(EGL_NO_X11)
-
-typedef void             *EGLNativeDisplayType;
-typedef khronos_uintptr_t EGLNativePixmapType;
-typedef khronos_uintptr_t EGLNativeWindowType;
-
-#elif defined(__unix__) || defined(USE_X11)
+#elif defined(USE_X11)
 
 /* X11 (tentative)  */
 #include <X11/Xlib.h>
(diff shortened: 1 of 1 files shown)


stonesthrow (approved):

stonesthrow:
Seems OK to me. @oddhack - how does this look?

oddhack:
> Seems OK to me. @oddhack - how does this look?

I think it would be good to get signoff from both Mesa and NVIDIA on this change - I am not sure which, if either, @kbrenneman represents here. The actual change seems fine to me but I don't use all those combinations of platforms.

stonesthrow:
@kbrenneman , can we get an review/ compile check/ reassurance from x11 community on this change?

cubanismo:
@kbrenneman is with NVIDIA. @nwnk for feedback from X.org side and @ianromanick because he objected when we tried to redefine EGLint one time, and this is vaguely reminiscent of that but much less invasive IMHO.

cubanismo (approved):
LGTM, but note @kbrenneman and I both represent NVIDIA.

stonesthrow:
Thanks James. I was just hoping for someone from x.org to poke for input. Just to be careful.

mattst88:
I'm very pleased to see this, with my Mesa and Gentoo/Linux hats on.

PR #132, opened 2021-08-14
extensions: add EGL_EXT_present_opaque

1ace:
This extension adds a new EGL surface attribute `EGL_PRESENT_OPAQUE_EXT` to indicate that the surface should be presented as opaque, disregarding any alpha channel if present.

/cc @cubanismo @fooishbar @flibitijibibo who were part of the discussion that lead to this extension.

checklist:
- [x] NVIDIA ack (for `EXT`)
- [ ] [Mesa implementation](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/12373)
- [ ] client implementation: [SDL](https://github.com/libsdl-org/SDL/pull/4646)

### DIFF
--- a/extensions/EXT/EGL_EXT_present_opaque.txt
+++ b/extensions/EXT/EGL_EXT_present_opaque.txt
@@ -1,0 +1,113 @@
+Name
+
+    EXT_present_opaque
+
+Name Strings
+
+    EGL_EXT_present_opaque
+
+Contributors
+
+    Eric Engestrom
+
+Contacts
+
+    Eric Engestrom (eric 'at' engestrom.ch)
+
+IP Status
+
+    No known IP claims.
+
+Status
+
+    Draft
+
+Version
+
+    #1, August 8, 2021
+
+Number
+
+    EGL Extension #146
+
+Extension Type
+
+    EGL display extension
+
+Dependencies
+
+    Requires EGL 1.4 and EGL_KHR_image_base extension
+
+    This extension is written against the wording of the EGL 1.4
+    specification, and EGL_KHR_image_base version 6.
+
+Overview
+
+    This extension adds a new EGL surface attribute EGL_PRESENT_OPAQUE_EXT
+    to indicate that the surface should be presented as opaque,
+    disregarding any alpha channel if present.
+    If surface attribute EGL_PRESENT_OPAQUE_EXT is EGL_TRUE, then the
+    surface will be presented as opaque.
+
+New Types
+
+    None
+
+New Procedures and Functions
+
+    None
+
+New Tokens
+
+    New EGLSurface attribute name:
+
+        EGL_PRESENT_OPAQUE_EXT                  0x31DF
+
+
+Additions to Chapter 3 of the EGL 1.4 Specification (Rendering Surfaces)
+
+    Change the second paragraph in section 3.5 on p. 28 (describing
+    eglCreateWindowSurface):
+
+        "Attributes that can be specified in attrib list include EGL_RENDER_BUFFER,
+        EGL_PRESENT_OPAQUE_EXT, EGL_VG_COLORSPACE, and EGL_VG_ALPHA_FORMAT."
+
+    Add the following paragraph in section 3.5 on p. 28 before
+    "EGL_VG_COLORSPACE specifies the color space used by OpenVG"
+    (describing eglCreateWindowSurface attrib_list):
+
+        "EGL_PRESENT_OPAQUE_EXT specifies the presentation opacity mode
+        of the window surface. If its value is EGL_TRUE, then the
+        surface's alpha channel (if any) will be ignored and considered
+        fully opaque. If its value is EGL_FALSE, then the surface's
+        alpha channel is respected by the compositor. The default value
+        of EGL_PRESENT_OPAQUE_EXT is EGL_FALSE."
+
+    Add to Table 3.5: Queryable surface attributes and types on p. 37
+
+        EGL_PRESENT_OPAQUE_EXT    boolean    Surface presentation opacity mode
+
+    Add following the second paragraph in section 3.6 on p. 39 (describing
+    eglQuerySurface):
+
+        "Querying EGL_PRESENT_OPAQUE_EXT returns the presentation
+        opacity mode of the surface. The presentation opacity mode of
+        window surfaces is specified in eglCreateWindowSurface. The
+        presentation opacity mode of pbuffer and pixmap surfaces is
+        always EGL_FALSE."
+
+    Add following after "which must be a valid native pixmap handle." in section 3.9.2 on
+    p. 53 (describing eglCopyBuffers):
+
+        "If attribute EGL_PRESENT_OPAQUE_EXT of surface has value of EGL_TRUE, then
+        an EGL_BAD_ACCESS error is returned."
+
+Issues
+
+    None
+
+Revision History
+
+    Version 1, 2021-08-08 (Eric Engestrom)
+      - Initial draft
+
--- a/extensions/EXT/EGL_EXT_present_opaque.txt
+++ b/extensions/EXT/EGL_EXT_present_opaque.txt
@@ -20,7 +20,7 @@ IP Status
 
 Status
 
-    Draft
+    Complete
 
 Version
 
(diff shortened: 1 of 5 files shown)


stonesthrow:
Why is this preferred over selecting a a no-alpha channel config, Or setting alpha to 1?

flibitijibibo:
> Why is this preferred over selecting a a no-alpha channel config, Or setting alpha to 1?

The main issue we ran into in SDL was the application requiring RGBA8888 surfaces (for example) but not wanting alpha to be presented as transparent - forcing the surface format to 0 causes issues if the format ends up being RGB888, rather than RGBX8888.

Two real-world examples are Portal 2 and Left 4 Dead 2; the latter is actually a more extreme example where they appear to use the alpha channel as some kind of special storage when rendering zombies:

https://twitter.com/flibitijibibo/status/1404888436184596486

This extension effectively gives us the equivalent of VkCompositeAlphaFlagBitsKHR.

flibitijibibo:
Prior discussion of alpha-by-default which is probably worth linking to directly:

https://gitlab.freedesktop.org/mesa/mesa/-/issues/4933#note_968244

stonesthrow:
Curiously - what is the expected implementation?

fooishbar:
> Curiously - what is the expected implementation?

Returning alpha-ful surfaces to the app, so it has full use of the alpha channel in winsys surfaces, however then instructing the winsys to use a constant alpha of 1.0 during presentation rather than sampling the alpha channel from the client surface.

This doesn't make a massive amount of sense on first reading, but it does fit with what you get from X11 most of the time, by virtue of X11 having native visuals which dictate the interpretation of alpha. Most apps will select a non-alphaful native visual, then create a surface with an alphaful config, which gives that behaviour.

Wayland has no equivalent to this since we don't have separate native visuals or formats: we just take the format of the buffer as created by the client. So if you create a surface with an alphaful config, that is passed on to the server and the alpha is treated as meaningful (i.e. blended) by the winsys. This flag gives us an opt-out so we can emulate X11 semantics and make porting easier.

extensions/EXT/EGL_EXT_present_opaque.txt
```diff
@@ -0,0 +1,109 @@
...
+
+Version
+
+    #1, August 8, 2021
+
+Number
+
+    EGL Extension #146
+
+Dependencies
```
stonesthrow, line 33: Extension Type

    EGL display extension

stonesthrow (commented):
Nit: add extension type

extensions/EXT/EGL_EXT_present_opaque.txt
```diff
@@ -0,0 +1,109 @@
...
+    Add the following paragraph in section 3.5 on p. 28 before
+    "EGL_VG_COLORSPACE specifies the color space used by OpenVG"
+    (describing eglCreateWindowSurface attrib_list):
+
+        "EGL_PRESENT_OPAQUE_EXT specifies the presentation opacity mode
+        of the window surface. If its value is EGL_TRUE, then the
+        surface's alpha channel (if any) will be ignored and considered
+        fully opaque. If its value is EGL_FALSE, then the surface's
+        alpha channel is respected by the compositor. The default value
+        of EGL_PRESENT_OPAQUE_EXT is EGL_FALSE."
```
cubanismo, line 80: I think what we would do is only advertise this on wayland-platform EGLDisplays, but is there any concern that when this extension is advertised, apps that set EGL_PRESENT_OPAQUE_EXT to EGL_FALSE (or leave it at the default) and rely on native winsys mechanisms to actually present transparent/translucent content will be violating the spec? Or is the spec intended to override the native mechanism in that case?

stonesthrow:
If Wayland specific, rename extension to EGL_WL_present_opaque ? + EGL_PRESENT_OPAQUE_WL.

cubanismo:
I don't know that this would always be Wayland-specific. That's just where we would likely implement it in our driver stack for now (Our Wayland support is a separate library external to our driver). We could probably implement it on our in-driver EGLOutput+EGLStream path too if there was demand, as it has roughly the same issue in theory, but I view this extension as something of a work-around/CYA bit (albeit a necessary one), so I don't want to propagate it too far if we can avoid it.

extensions/EXT/EGL_EXT_present_opaque.txt
fooishbar: I'm not sure what you mean here - can you please reword?

1ace:
Sorry I wasn't able to respond yesterday, but I see that others have explained things already.

I don't want to repeat too much, but this extension is indeed about letting an app do whatever it wants with the alpha channel, as a lot of big games apparently do, while being able to present it to a compositor without ending up with whatever was behind the app showing through.

extensions/EXT/EGL_EXT_present_opaque.txt
1ace: I don't quite understand your question either, but I tried to word it as:
> `EGL_FALSE` => winsys behaviour is unchanged (ie. if it normally ignores the alpha channel it will keep doing that, and if it normally uses it it wil keep doing that

and
> `EGL_TRUE` => winsys pretends alpha is 1 (or whatever value "opaque" is)

I'm happy to change the text if you think that's not clear :)

fooishbar: FALSE is the default. There is no conformant Wayland implementation that acts as if it were true.

Given the influence of visuals and composition, I don’t think this could be made to work at all correctly under X11. It might be relevant for EGLOutput to display directly I guess?

cubanismo: Sorry if I wasn't clear. It is indeed the X11  case I had in mind. If we exposed this extension on X11, I don't know how we would implement it. If the app set EGL_PRESENT_OPAQUE_EXT to EGL_TRUE, we could hypothetically override SwapBuffers to write 1.0 to the alpha channel, and EGL doesn't have front-buffered rendering, so that would solve it, but it's unfortunate to need that invasive of a change for this simple extension.

Regardless, what I'm more worried about is an app selecting an X visual that has no alpha channel, then setting this to EGL_FALSE, or more realistically, leaving it at the default. My reading of the above sentence, "If its value is EGL_FALSE, then the surface's alpha channel is respected by the compositor," was that the compositor must use the alpha channel provided by the app. However, the compositor doesn't talk to the app directly, nor vice-versa, so it has to rely on the window's visual to determine whether there is an alpha channel, and that's going to say there isn't one, so it will behave as if EGL_PRESENT_OPAQUE_EXT were set to EGL_TRUE.

The plan is that we're just not going to advertise this for X11 apps, so perhaps the outcome is just academic. I was just curious whether others planned to advertise this on X11 or similar environments, and hence were interesting in a more relaxed wording corresponding to what you list as your intent above, that EGL_FALSE == status quo.

fooishbar: > The plan is that we're just not going to advertise this for X11 apps, so perhaps the outcome is just academic. 

Yeah, I don’t think you can. On X11 this flag is effectively determined by the native visual of the config you select - which isn’t a concept we have on Wayland, hence why we need to provide something new to give apps the same behaviour they get by applying an alphaful config to a non-alphaful visual …

> we could hypothetically override SwapBuffers to write 1.0 to the alpha channel

… which is why I don’t think this could work, because in the presence of buffer_age you would’ve destroyed visible and meaningful data.

In both cases (alphaful config with either Wayland and this property, or X11 and a non-alpha visual) we get the same behaviour: client gets to write to and read back from the alpha channel of the default framebuffer, but the presentation engine treats it as if it were constant 1.0.

1ace: How about this wording?
> If its value is EGL_FALSE, then the compositor doesn't change its behaviour, and considers the surface's alpha channel the same way as if the extension wasn't implemented.

And yeah, for Mesa I don't think we'll have any reason to implement this on our other platforms than Wayland; you'll probably want to do the same in the NVIDIA driver.

cubanismo: That'll work, yeah.

fooishbar: Wording WFM.

fooishbar:
> but I view this extension as something of a work-around/CYA bit (albeit a necessary one), so I don't want to propagate it too far if we can avoid it.

Believe me, it's not that I _want_ to implement this ...

cubanismo (approved):
LGTM.

stonesthrow (approved):
Extension form seem good and complete

stonesthrow:
@1ace , You have the NVidia task complete. When all ready and we have a 3rd approve, I will move to "ready for merge"

1ace:
Perhaps @fooishbar and/or @flibitijibibo can give their approval, representing Mesa which will have to provide this extension, and SDL which will be using it.

(both implementations are done already, but this is to make sure that they are ok with "what" these do; "how" can be reviewed in their respective merge/pull requests)

flibitijibibo (approved):
Looks great!

stonesthrow:
With the approvals, and 1ace says there are implementations, I'm marking this "approved to merge"

PR #134, opened 2021-09-01
Reserve enum for EGL_OPENWF_DEVICE_EXT

maniraj87:
With revision #5 of EGL_EXT_device_openwf extension, EGL_OPENWF_DEVICE_EXT
attribute was added to the spec but it was somehow not added in egl.xml,
adding it now.


stonesthrow:
Looks to be in the middle of a large NVidia bloc of reserved enums. But looks like that's what we have done with OpenWF.
No objections.

cubanismo (approved):

stonesthrow (approved):

PR #137, opened 2021-10-06
Use (u)intptr_t in khrplatform.h if available

arichardson:
While most architectures allow casting a pointer to `long`/`long long`
and back, this does not work on CHERI-enabled architectures such as
e.g. Arm Morello, where pointers are 128-bits but `long long` is still 64.
We use (u)inptr_t when possible to handle more platforms. Ideally, we
could just use (u)intptr_t everywhere, but this could result in ABI
breakage if khronos_uintptr_t is changed from unsigned long to
unsigned long long or similar (this results in different C++ name
mangling). To avoid changes for existing platforms, we restrict usage
of intptr_t to platforms where the size of a pointer is larger than the
size of long.

### DIFF
--- a/api/KHR/khrplatform.h
+++ b/api/KHR/khrplatform.h
@@ -153,6 +153,20 @@ typedef int64_t                 khronos_int64_t;
 typedef uint64_t                khronos_uint64_t;
 #define KHRONOS_SUPPORT_INT64   1
 #define KHRONOS_SUPPORT_FLOAT   1
+/*
+ * To support platform where unsigned long cannot be used interchangeably with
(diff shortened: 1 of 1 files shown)


arichardson:
> This will impact EGL, OpenGL ES, Open GL. So I'll ask that they check this over just to be certain. @pdaniell-nv , @cubanismo

Thanks,

It is possible there could be some C++ name mangling changes for platforms that define uintptr_t as `unsigned long long` instead of `unsigned long`.

cubanismo:
Hmm, an ABI change does sound pretty risky. I suppose the alternative is yet another ifdef to handle this particular platform correctly?

pdaniell-nv:
Approved for merge by the OpenGL/ES working groups on 2021-11-10.

PR #141, opened 2021-11-17
Adding EGL_EXT_surface_compression

janharaldfredriksen-arm:


janharaldfredriksen-arm:
CC: @pdaniell-nv, @oddhack, @stonesthrow

extensions/EXT/EGL_EXT_surface_compression.txt
```diff
@@ -0,0 +1,289 @@
...
+
+New Tokens
+
+    New attributes accepted by the <attrib_list> argument of
+    eglCreatePlatformWindowSurface and eglCreateWindowSurface:
+        EGL_SURFACE_COMPRESSION_EXT                     0x34B0
+
+    [Only if EGL_EXT_yuv_surface is supported]
+        EGL_SURFACE_COMPRESSION_PLANE1_EXT              0x328E
+        EGL_SURFACE_COMPRESSION_PLANE2_EXT              0x328F
```
stonesthrow, line 72: You do not expect 3 plane YUV?

extensions/EXT/EGL_EXT_surface_compression.txt
```diff
@@ -0,0 +1,289 @@
...
+
+    Accepted as attribute values for EGL_SURFACE_COMPRESSION_EXT by
+    eglCreatePlatformWindowSurface and eglCreateWindowSurface:
+        EGL_SURFACE_COMPRESSION_FIXED_RATE_NONE_EXT     0x34B1
+        EGL_SURFACE_COMPRESSION_FIXED_RATE_DEFAULT_EXT  0x34B2
+
+        EGL_SURFACE_COMPRESSION_FIXED_RATE_1BPC_EXT     0x34B4
```
stonesthrow, line 79: This is very specific to me. What is the benefit to the application to know the compression rate?
Lets take for example ASTC - you get a a compression block, it tells you some information about format. Which might be something that is compatible with Window System or not. So that is my question rate vs format, seems format would be more useful. And would that then lead to having configs so you can choose a compression format? Or is this YUV specific compression on top of a YUV format? How is that rate passed to windows system? decoded from buffer?

extensions/EXT/EGL_EXT_surface_compression.txt
```diff
@@ -0,0 +1,289 @@
...
+    Querying EGL_SURFACE_COMPRESSION_PLANE2_EXT returns the actual
+    fixed-rate compression applied to plane 2 of a YUV surface.
+    These values may be different to the one requested when the surface was created.
+    For pbuffer and pixmap surfaces, the contents of <value> are not modified."
+
+    To get the list of all fixed-rate compression rates that are available on
+    a specified display and EGLConfig, call
+
+       EGLBoolean eglQuerySupportedCompressionRatesEXT(
+           EGLDisplay dpy, EGLConfig config, const EGLAttrib *attrib_list,
```
stonesthrow, line 209: config is the format, rates are compression options on top of that? And window System can decode the compression from the buffer?

stonesthrow (commented):
Just some question to understand this

extensions/EXT/EGL_EXT_surface_compression.txt
stonesthrow: ACK.

extensions/EXT/EGL_EXT_surface_compression.txt
janharaldfredriksen-arm: There are some long discussion on this in the Khronos internal issues. 
New formats was an option, but lead to a very large number of new formats - as most vendors support a small handful of compression rates per format.
How the rate is passed around is somewhat an implementation detail, but could be done with private Gralloc flags or DRM format modifiers.

extensions/EXT/EGL_EXT_surface_compression.txt
janharaldfredriksen-arm: Yes. (Some overlap with the reply to the previous question.)

extensions/EXT/EGL_EXT_surface_compression.txt
stonesthrow: The buffer its self would be un compressed size then, And app will pick compression. 
It reminds me a bit of the old BAUD rate for dial up modems - lots of rates 9600 ->38k. But I think in use, wont an app either pick no compression, or maximum compression? I'm just not sure an app/developer/user would choose one over another with so many choices. How would that be selected or determined? Just curios. Or in practice, would you expect a limited set?

janharaldfredriksen-arm: Yes, in practice, each implementation will most likely support a subset of bitrates per format. For our implementation, it's 3 rates per format. eglQuerySupportedCompressionRatesEXT will return the rates the given implementation actually supports.
We do provide the EGL_SURFACE_COMPRESSION_FIXED_RATE_DEFAULT_EXT option that developers can use if they just want the implementation to pick a "best" compression rates. The specific rates are there for anyone wanting to fine-tune beyond that (and also that we can allow queries to return some information of what compression rate you actually got).

cubanismo: This seems like it should all be covered by DRM format modifiers and application or vendor-specific logic layered on top of it as desired. Why is a further extension needed/useful, and how would the two interact?

janharaldfredriksen-arm: Our implementation will use this information - together with various internal state - to decide what DRM format modifiers to apply. The DRM format modifiers will be vendor specific. We'll likely make different decisions for different generations of hardware.
I can't speak to what other vendors plan to do.
In that sense, this API acts as an abstraction across vendors and HW generations. This also keeps the selection mechanism the same between this and the corresponding GLES extension, which also applies to driver-internal images.

stonesthrow: Thanks

janharaldfredriksen-arm:
@stonesthrow - if you're happy with the responses above, can we merge this?

stonesthrow:
Its ready when you say it is.

extensions/EXT/EGL_EXT_surface_compression.txt
```diff
@@ -0,0 +1,289 @@
...
+
+    EGLBoolean eglQuerySupportedCompressionRatesEXT(
+           EGLDisplay dpy, EGLConfig config, const EGLAttrib *attrib_list,
+           EGLint *rates, EGLint rate_size, EGLint *num_rates);
+
+New Tokens
+
+    New attributes accepted by the <attrib_list> argument of
+    eglCreatePlatformWindowSurface and eglCreateWindowSurface:
+        EGL_SURFACE_COMPRESSION_EXT                     0x34B0
```
oddhack, line 68: The extension itself looks fine, but you will also need to add it to egl.xml before we merge this, including the new entry points and enums as well as the new \<extension> block; and also to registry.tcl so the index gets regenerated properly.

oddhack (changes requested):

oddhack:
Assigned back to @janharaldfredriksen-arm until egl.xml / registry.tcl updates are included

stonesthrow:
Thanks Jon. Sorry I didn't catch. Too preoccupied.

janharaldfredriksen-arm:
@oddhack - apologies. I had done the XML updates but obviously forgotten to git add them.

They should be included now. (I've also included the generated headers, but can obviously back those out if you prefer.)

oddhack (approved):
Thanks, LGTM. Merging.

PR #145, opened 2021-12-06
Add EGL_PROTECTED_CONTENT_EXT queries for Surface and Context

stonesthrow:
Added EGL_PROTECTED_CONTENT_EXT queries for eglQuerySurface and eglQueryContext
This satisfies some testing purposes.

### DIFF
--- a/extensions/EXT/EGL_EXT_protected_content.txt
+++ b/extensions/EXT/EGL_EXT_protected_content.txt
@@ -82,15 +82,12 @@ New Procedures and Functions
 
 New Tokens
 
-    New EGLSurface attribute name:
+    Accepted as an attribute name in the <attrib_list> parameter of
+    eglCreateContext, eglCreateWindowSurface, and eglCreateImageKHR;
+    and as an parameter of eglQuerySurface and eglQueryContext:
 
         EGL_PROTECTED_CONTENT_EXT               0x32C0
         
-    Accepted as an <attribute> parameter of eglQuerySurface and 
-    eglQueryContext:
-
-        EGL_PROTECTED_CONTENT_EXT               0x32C0
-
 
 Add a new section 2.7 entitled "Protected Content" at the end of Chapter 2 (EGL
 Operation)
--- a/extensions/EXT/EGL_EXT_protected_content.txt
+++ b/extensions/EXT/EGL_EXT_protected_content.txt
@@ -34,7 +34,7 @@ Status
 
 Version
 
-    Version 12, April 14, 2016
+    Version 13, December 6, 2021
 
 Number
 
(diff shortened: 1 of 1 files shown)


stonesthrow:
reference: https://gitlab.khronos.org/egl/Documents/-/issues/10

extensions/EXT/EGL_EXT_protected_content.txt
```diff
@@ -85,6 +85,12 @@ New Tokens
     New EGLSurface attribute name:
 
         EGL_PROTECTED_CONTENT_EXT               0x32C0
+        
+    Accepted as an <attribute> parameter of eglQuerySurface and
```
oddhack, line 89: These could be combined rather than repeat the enum definition twice - something like

> Accepted as an attribute name in the <attrib_list> parameter of eglCreateContext, eglCreateWindow Surface, and eglCreateImageKHR; and as an <attribute> parameter of eglQuerySurface and eglQueryContext:

oddhack:
This could refer to EGL 1.5 instead of 1.4, in which some of the extension interfaces referred to have been promoted to core. Probably not worth the effort to renumber all the section references, though.

I am slightly concerned that adding interfaces to an extension that's been out for 5+ years could cause problems with shipping implementations, but haven't followed the development of this update so maybe you've considered that?

stonesthrow:
Reply to @oddhack , Agreed it is not good to add more requirements to a old spec. We did check IHVs to see who implemented and who did not - reference link above. Most did already, and others said they would. There is a Android test that uses this, and that is the initiator. Thank you for feedback. I'll push the other way too, and see if Google test can be made compliant to surrent spec..

extensions/EXT/EGL_EXT_protected_content.txt
stonesthrow: Yes, better

stonesthrow: Done

oddhack:
LGTM, thanks.

extensions/EXT/EGL_EXT_protected_content.txt
```diff
@@ -200,6 +211,11 @@ Additions to Chapter 3 of the EGL 1.4 Specification (Rendering Surfaces)
 
         The default value of EGL_PROTECTED_CONTENT_EXT is EGL_FALSE."
 
+    Add the following paragraph in section 3.7.4 Context Queries. Add after
+    the last paragraph after eglQueryContext queries.
+    
+        "Querying EGL_PROTECTED_CONTENT_EXT returns the current value"
+
```
gnl21, line 215: These additions look like they are the wrong way round in the PR to me. That is, the addition to section 3.5.6 follows the part about adding to section 3.7 and this one, which adds to section 3.7.4 follows the part about adding to section 3.5. (I don't know why the original paragraphs were in reverse order).

stonesthrow: OK. I will check that. thx.

gnl21: Thanks. Apart from that, I've reviewed this and I'm fine with it.

stonesthrow: I sure did get them mixed. In a rush I guess. thanks - fixed.

pdaniell-nv:
We discussed this on 2021-12-08 and agree to approve it in a couple of weeks if there were no objections. Setting this as approved to merge.

stonesthrow:
Thank you!

stonesthrow:
@oddhack - please merge

Issue #146, opened 2021-12-09
[EGL-Registry mismatch with EGL spec] Why the default value of EGL_GL_COLORSPACE on eglCreatePBufferSurface has gap between EGL spec with this EGL-Registry ?


JohnBro:
Hi KhronosGroup team,

we found that this EGL-Registry may have some mistake with EGL spec,
In this Registry, Regards to [eglCreatePbufferSurface.xhtml](https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglCreatePbufferSurface.xhtml),

> Description
> eglCreatePbufferSurface creates an off-screen pixel buffer surface and returns its handle. If eglCreatePbufferSurface fails to create a pixel buffer surface, EGL_NO_SURFACE is returned.
>
> Surface attributes are specified as a list of attribute-value pairs, terminated with EGL_NONE. Accepted attributes are:
> 
> EGL_GL_COLORSPACE
> Specifies the color space used by OpenGL and OpenGL ES when rendering to the surface. If its value is EGL_GL_COLORSPACE_SRGB, then a non-linear, perceptually uniform color space is assumed, with a corresponding GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING value of GL_SRGB. If its value is EGL_GL_COLORSPACE_LINEAR, then a linear color space is assumed, with a corresponding GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING value of GL_LINEAR. The default value of EGL_GL_COLORSPACE is EGL_GL_COLORSPACE_SRGB.

We can see the default value of EGL_GL_COLORSPACE is EGL_GL_COLORSPACE_SRGB.

But according to EGL-1.5 spec:
https://www.khronos.org/registry/EGL/specs/eglspec.1.5.pdf
Got info from eglCreatePlatformWindowSurface
![image](https://user-images.githubusercontent.com/31929573/145317507-16fcdf25-c3aa-4e68-b53b-5fdc995a8a1d.png)
Got info from eglCreatePbufferSurface
![image](https://user-images.githubusercontent.com/31929573/145317519-22dc68b1-a8db-43c0-bc1e-c08980f361c0.png)

So, could you please help to check this gap and make an alignment for EGL-Registry with EGL spec ？ Many thanks.

Best Regards,
Junbo

stonesthrow:
Thank you for pointing this out. We'll get it fixed. default colorspace is EGL_GL_LINEAR.

[...]

PR #147, opened 2021-12-09
Update eglCreatePbufferSurface.xhtml

stonesthrow:
The default color space is EGL_GL_COLORSPACE_LINEAR not SRGB


stonesthrow:
See: https://github.com/KhronosGroup/EGL-Registry/issues/146

oddhack:
LGTM. Do you need me to regenerate the HTML? I should probably create a Docker image with the docbook toolchain installed.

stonesthrow:
If it looks good to you then all in your hands to do all the steps needed to update. thanks

oddhack:
Ah, realized I had this backwards - the source is in ../eglCreatePbufferSurface.xml and the .xhtml is regenerated from that. I will fix the source as well.

PR #148, opened 2021-12-10
Add missing const keywords to eglStreamImageConsumerConnectNV

kbrenneman:
The `modifiers` and `attrib_list` parameters to `eglStreamImageConsumerConnectNV` in EGL_NV_stream_consumer_eglimage are supposed to be input parameters, not output, but they're missing the const keyword.

This fixes the extension spec and the XML/header files.

Fixing the definition in eglext.h should be safe to change without breaking anything, since you can always pass a non-const value as a const parameter.

### DIFF
--- a/api/EGL/eglext.h
+++ b/api/EGL/eglext.h
@@ -14,12 +14,12 @@ extern "C" {
 ** used to make the header, and the header can be found at
 **   http://www.khronos.org/registry/egl
 **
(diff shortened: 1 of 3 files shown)


stonesthrow (approved):
Seems right

oddhack (approved):
LGTM. @stonesthrow please set the Approve to Merge milestone and I'll merge it.

PR #153, opened 2021-12-15
Correct and clarify EGLSync text in EGL_NV_stream_consumer_eglimage

kbrenneman:
This is another correction for EGL_NV_stream_consumer_eglimage.

The text currently says to use "a reusable EGLSync object", which is ambiguous, because you've got both reusable syncs (EGL_KHR_reusable_sync) and reusable *fence* syncs (EGL_EXT_sync_reuse).

The extension actually needs a fence sync -- that is, a sync created with type `EGL_SYNC_FENCE`, but created in a signaled state. Passing in an EGLSync of type `EGL_SYNC_REUSABLE_KHR` doesn't work (and doesn't really make sense).

Likewise, the "Dependencies" section is incorrect, because it should require EGL_EXT_sync_reuse, not EGL_KHR_reusable_sync.

This change corrects the dependency text, and rewrites the descriptions of `eglStreamAcquireImageNV` and `eglStreamReleaseImageNV` to (hopefully) make it more clear what you're supposed to do with the EGLSync parameters, and to make the error codes explicit.

Note that this PR also includes my previous change from #148, so that it should merge cleanly after #148 is merged. I can separate it out and then rebase it later if that makes things easier, though.

### DIFF
--- a/extensions/NV/EGL_NV_stream_consumer_eglimage.txt
+++ b/extensions/NV/EGL_NV_stream_consumer_eglimage.txt
@@ -39,7 +39,7 @@ Dependencies
 
     Requires the EGL_KHR_stream extension.
 
-    Requires the EGL_KHR_reusable_sync extension.
+    Requires the EGL_EXT_sync_reuse extension.
 
     This extension is written against the wording of the EGL 1.5
     Specification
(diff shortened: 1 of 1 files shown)


stonesthrow (approved):
LGTM

kbrenneman:
I just rebased the change so that it'll merge cleanly.

oddhack:
@stonesthrow if you want to set Approved to Merge on this I'll do it, or you can just merge it - doesn't require anything else to get auto-propagated to the khronos.org registry page.

PR #156, opened 2022-03-25
Add EGL_EXT_display_alloc extension

kbrenneman:
This is an EGL extension that allows you to allocate private EGLDisplays and destroy them when they're no longer needed.

The default semantics of eglGetPlatformDisplay have a few inherent problems:

First, if two modules (applications, libraries, etc) both try to use EGL internally, then they have to sharing the same EGLDisplay, and they risk clobbering each other. `EGL_KHR_display_reference` can at least make the problem less catastrophic when something calls `eglTerminate`, but it's not a complete solution -- an extra call to `eglTerminate` would still break everything, and you can't rely on using eglTerminate to clean up any remaining display-owned objects.

Likewise, the fact that EGLDisplay handles stick around forever means that it's literally impossible to write an EGL application (or implementation) that doesn't leak memory.

Perhaps most importantly, the default semantics make no sense at all in X11 or Wayland applications, because the EGLDisplay will necessarily last longer than the display connection under it, leading to a dangling pointer.

So, this extension adds a new EGL_PRIVATE_DISPLAY_EXT attribute for eglGetPlatformDIsplay, which tells it to allocate a brand new EGLDisplay handle, even if another EGLDisplay would otherwise match the native display. Different modules can thus use their own private EGLDisplays without risk of clobbering each other.

It also adds a new eglDestroyDisplayEXT function, which will fully destroy an EGLDisplay handle that was created with EGL_PRIVATE_DISPLAY_EXT. That way, you can actually clean up properly, or deal with cases where a native display is no longer valid.

This would also make it possible to implement new client extensions such as EGL_KHR_display_reference entirely within libglvnd.

### DIFF
--- a/api/egl.xml
+++ b/api/egl.xml
@@ -886,7 +886,8 @@
         <enum value="0x3376" name="EGL_STREAM_IMAGE_AVAILABLE_NV"/>
         <enum value="0x3377" name="EGL_DRM_RENDER_NODE_FILE_EXT"/>
         <enum value="0x3378" name="EGL_STREAM_CONSUMER_IMAGE_USE_SCANOUT_NV" />
-            <unused start="0x3379" end="0x339F"/>
+        <enum value="0x3379" name="EGL_ALLOC_NEW_DISPLAY_EXT"/>
+            <unused start="0x337A" end="0x339F"/>
     </enums>
 
     <enums namespace="EGL" start="0x33A0" end="0x33AF" vendor="ANGLE" comment="Reserved for Shannon Woods (Bug 13175)">
@@ -2066,6 +2067,10 @@
              <param><ptype>EGLint</ptype> <name>max_size</name></param>
              <param>void *<name>value</name></param>
              <param>EGLint *<name>size</name></param>
+        </command>
+        <command>
+             <proto><ptype>EGLBoolean</ptype> <name>eglDestroyDisplayEXT</name></proto>
+             <param><ptype>EGLDisplay</ptype> <name>dpy</name></param>
         </command>
     </commands>
 
@@ -3578,6 +3583,12 @@
         <extension name="EGL_QNX_image_native_buffer" supported="egl">
             <require>
                 <enum name="EGL_NATIVE_BUFFER_QNX"/>
+            </require>
+        </extension>
+        <extension name="EGL_EXT_display_alloc" supported="egl">
+            <require>
+                <enum name="EGL_ALLOC_NEW_DISPLAY_EXT"/>
+                <command name="eglDestroyDisplayEXT"/>
             </require>
         </extension>
     </extensions>
--- a/registry.tcl
+++ b/registry.tcl
@@ -793,4 +793,9 @@ extension EGL_ANDROID_telemetry_hint {
     flags       public
     filename    extensions/EXT/EGL_ANDROID_telemetry_hint.txt
 }
-# Next free extension number: 154
+extension EGL_EXT_display_alloc {
+    number      154
+    flags       public
+    filename    extensions/EXT/EGL_EXT_display_alloc.txt
+}
+# Next free extension number: 155
--- a/extensions/EXT/EGL_EXT_display_alloc.txt
+++ b/extensions/EXT/EGL_EXT_display_alloc.txt
@@ -1,0 +1,193 @@
+Name
+
+    EXT_display_alloc
+
+Name Strings
+
+    EGL_EXT_display_alloc
+
+Contributors
+
+Contacts
+
+    Kyle Brenneman,  NVIDIA  (kbrenneman 'at' nvidia.com)
+
+Status
+
+    Complete.
+
+Version
+
+    Version 1 - March 25, 2022
+
+Number
+
+    EGL Extension #154
+
+Extension Type
+
+    EGL client extension
+
+Dependencies
+
+    Written based on the wording of the EGL 1.5 specification.
+
+    Requires EGL_EXT_platform_base or EGL 1.5
+
+    Interacts with the EGL_KHR_display_reference extension
+
+    This extension includes the function eglQueryDisplayAttribEXT as
+    defined in EGL_EXT_device_query.
+
+Overview
+
+    The existing semantics of EGLDisplay object lifetimes work well for
+    applications in which one module manages all EGL usage, and in which
+    EGL displays and the underlying native displays are expected to
+    remain available until application termination once they are
+    instantiated. However, EGL does not provide reasonable semantics for
+    the following use cases:
+
+      1) Applications that rely on toolkit libraries which use EGL
+         independently from the application itself.
+
+      2) Applications that use multiple native display objects with EGL
+         and wish to completely release resources associated with
+         EGLDisplay objects they no longer need.
+
+      3) Platforms where the native display connection can be closed,
+         such as X11 or Wayland. If the native display is closed, then
+         there's no possible reasonable behavior for an EGLDisplay.
+
+    This extension solves these problems by adding the ability to create
+    multiple unique, unshared EGLDisplay objects from the same native
+    display, and to destroy those EGLDisplay objects when the caller no
+    longer needs them.
+
+New Types
+
+    None
+
+New Functions
+
+    EGLBoolean eglDestroyDisplayEXT(EGLDisplay dpy);
+
+New Tokens
+
+    Accepted as an attribute in the <attrib_list> parameter of
+    eglGetPlatformDisplay and the <name> parameter of
+    eglQueryDisplayAttribEXT:
+
+        EGL_ALLOC_NEW_DISPLAY_EXT                       0x3379
+        
+In section "3.2 Initialization":
+
+Remove the sentence in the description of eglGetPlatformDisplay
+indicating no valid attribute names are defined, and add the following:
+
+    If the EGL_ALLOC_NEW_DISPLAY_EXT attribute is set to EGL_TRUE, then
+    eglGetPlatformDisplay creates a new, unique EGLDisplay handle, even
+    if the platform, native_display, and attributes would match an
+    existing EGLDisplay.
+
+    The default value of the EGL_ALLOC_NEW_DISPLAY_EXT attribute is
+    EGL_FALSE.
+
+Add to the end of section "3.2 Initialization":
+
+    To destroy an EGLDisplay handle, use:
+
+        EGLBoolean eglDestroyDisplayEXT(EGLDisplay dpy);
+
+    eglDestroyDisplayEXT frees the handle <dpy>.
+
+    If <dpy> is initialized, or if it is current to one or more threads,
+    then is is marked for deletion, and it will be destroyed when it is
+    terminated and no longer current.
+
+    After <dpy> is destroyed, it is not valid to pass to any EGL
+    function. Subsequent calls to eglGetPlatformDisplay may re-use the
+    EGLDisplay handle value for new displays.
+
+    If <dpy> was not created with the EGL_ALLOC_NEW_DISPLAY_EXT attribute
+    set to EGL_TRUE, then an EGL_BAD_ACCESS error is generated.
+
+Changes to section "3.3 EGL Queries":
+
+    Add EGL_ALLOC_NEW_DISPLAY_EXT to the attributes accepted by
+    eglQueryDisplayAttribEXT.
+
+Interactions with EGL_KHR_display_reference:
+
+    If a display has both EGL_ALLOC_NEW_DISPLAY_EXT and
+    EGL_TRACK_REFERENCES_KHR set to EGL_TRUE, then a call to
+    eglDestroyDisplayEXT will destroy the display after the display's
+    initialization count reaches zero.
+
+    That is, the display handle is not destroyed until eglTerminate has
+    been called as many times as eglInitialize.
+
+Issues
+
+    1.  What happens if an EGLDisplay is deleted while it's current to a
+        thread?
+
+        RESOLVED: Follow the same behavior as destroying a current
+        EGLContext or EGLSurface. Mark the display for deletion, but
+        don't delete it until it's no longer current.
+
+    2.  What happens if an EGLDisplay is deleted while it's initialized?
+        How should this interact with EGL_KHR_display_reference?
+
+        RESOLVED: An initialized display counts as "in use" for the
+        purposes of destruction, so the display doesn't get destroyed
+        until it's terminated. With EGL_KHR_display_reference, that just
+        means you have to call eglTerminate enough times to actually
+        terminate the display.
+
+        Treating an initialized display as "in use" allows for
+        compatibility with any libraries that get an EGLDisplay handle
+        from somewhere else, but expect to use EGL_KHR_display_reference
+        to maintain a reference to that display.
+
+        In addition, that keeps eglDestroyDisplayEXT and eglTerminate
+        orthogonal. If eglDestroyDisplayEXT destroyed an initialized
+        EGLDisplay, then it would effectively have one (or more)
+        implicit calls to eglTerminate.
+
+        Also note that unlike EGLContext and EGLSurface objects (whose
+        handles are invalidated immediately when they're marked for
+        deletion), an EGLDisplay handle remain valid until it's no
+        longer in use. That's necessary because eglMakeCurrent doesn't
+        accept EGL_NO_DISPLAY even to release the current context.
+
+    3.  Do we need a separate eglDestroyDisplayEXT function, instead of
+        just destroying the display when it's terminated?
+
+        RESOLVED: Yes. Using a separate function means that the
+        semantics of eglTerminate are unchanged. That keeps this
+        extension orthogonal to EGL_KHR_display_reference.
+
+    4.  What happens if you create two EGLDisplays from the same native
+        display?
+
+        RESOLVED: The two EGLDisplays will share the same native display
+        connection, but they are otherwise separate. Each EGLDisplay can
+        be initialized, terminated, and destroyed independently of one
+        another, and they're separate namespaces for EGLContexts,
+        EGLSurfaces, and other EGL objects.
+
+        Note that this not a change from core EGL: Calling
+        eglGetPlatformDisplay with the same native display but different
+        attribute lists already creates multiple EGLDisplay objects.
+
+        EGL_ALLOC_NEW_DISPLAY_EXT just tells eglGetPlatformDisplay to
+        create a new EGLDisplay even if the attribute lists are the
+        same.
+
+Revision History
+
+    #1 (March 25, 2022) Kyle Brenneman
+
+        - Initial draft
+


stonesthrow:
https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_display_reference.txt - is not sufficient then?
We had a lot of discussion at one time for that.
Just curious, who is part of this EXT?

kbrenneman:
EGL_KHR_display_reference is not sufficient, no. In terms of utility, it's effectively a strict subset of EGL_EXT_display_alloc.

With EGL_KHR_display_reference, you can have multiple libraries that call eglInitialize and eglTerminate independently without clobbering each other, which is arguably the worst problem with EGL's normal semantics. That also requires that you don't have any bugs that would lead to an extra eglTerminate call, and that each library cleans up every display-owned resource manually.

With EGL_EXT_display_alloc, multiple libraries can still call eglInitialize and eglTerminate independently because they would operate on separate EGLDisplay handles. If a library has an extra eglTerminate call, it doesn't matter. And a library could use eglTerminate to handle cleanup.

EGL_KHR_display_reference also doesn't fix the fact that everything that ever uses EGL has to leak memory, and it doesn't fix the dangling pointers when you call XCloseDisplay or wl_display_disconnect.

Perhaps most interestingly, EGL_EXT_display_alloc would make it possible to implement extensions like EGL_KHR_display_reference entirely within libglvnd. Libglvnd could implement reference tracking for an EGLDisplay internally pretty easily, but without EGL_EXT_display_alloc, it couldn't get separate EGLDisplay handles for different values of EGL_TRACK_REFERENCES_KHR.

As for the use of EXT, there's nothing hardware- or vendor-specific about this. It's just a change in a driver's bookkeeping, so any driver could implement it. I did float the general idea in a discussion a while ago with favorable response from someone on the Mesa side, though I don't remember who -- @nwnk maybe?

stonesthrow:
Sorry my comment was not push back. I got a better read and I get it.

stonesthrow (commented):
Comments.
Seems OK to me.

stonesthrow:
Not sure if rebasing on other work in progress has turned out without issue. Not sure.

kbrenneman:
Okay. Easy enough to just wait for #156 to merge and then rebase on ToT, then.

oddhack:
@kbrenneman you don't need to rebase, just please take out the generated files (eglext.h / index.php) to avoid conflicts. I will fix any XML conflicts (unlikely) and extension number issues prior to merging.

registry.tcl
```diff
@@ -762,4 +762,9 @@ extension EGL_EXT_surface_compression {
     flags       public
     filename    extensions/EXT/EGL_EXT_surface_compression.txt
 }
+extension EGL_EXT_display_alloc {
+    number      148
```
oddhack, line 766: This is going to end up as extension 149 (if it's the next one that is merged). Actually I guess egl.xml / registry.tcl do need to be rebased - I had forgotten that adding to the end of the file always generates conflicts between two branches.

stonesthrow:
Please add other reviewers. Jan and I can says its well formed, but +1 from experts is reassuring.

hamarb123:
I was looking to implement similar functionality to this - I want to be able to force a new allocation of a display. I would want my one to be destroyed the normal way using `eglTerminate`, and interact with `EGL_KHR_display_reference` to have additional logic for deleting. I also do not fully understand the purpose of `eglDestroyDisplayEXT` - is this just to provide different behaviour depending on `EGL_KHR_display_reference` - if so, would it be more logical to only require it be included it when both extensions are present? Would this extension meet my wants? Should the prefix be different to PRIVATE since I also want to use it?

kbrenneman:
In practice, I expect that most applications and libraries would use `EGL_EXT_display_alloc` instead of `EGL_KHR_display_reference`, not use the two of them together. The main purpose of `EGL_KHR_display_reference` is so that two libraries which end up getting the same EGLDisplay when they call `eglGetPlatformDisplay` won't clobber each other when one of them calls `eglTerminate` (assuming no bugs in either, of course). With `EGL_EXT_display_alloc`, you can simply not share an EGLDisplay in the first place.

The reason I added a separate `eglDestroyDisplayEXT` function is so that `eglInitialize` and `eglTerminate` would *not* have different behavior. Once you've got an EGLDisplay handle, calling `eglInitialize` and `eglTerminate` would do exactly the same thing regardless of whether you set `EGL_PRIVATE_DISPLAY_EXT` or not, including being able to reinitialize an EGLDisplay.

That also means that this extension is logically orthogonal to `EGL_KHR_display_reference`. This extension only cares about whether an EGLDisplay is initialized, not how many times you'd have to call `eglTerminate` to uninitialize it.

In addition, you can easily get the behavior that you want, too: Just call `eglDestroyDisplayEXT` immediately after `eglInitialize`. It'll mark the EGLDisplay for deletion, but since it treats an initialized EGLDisplay is "in use", it'll wait until `eglTerminate` to actually destroy it.

hamarb123:
Cool, sounds good.
Just checking, the difference between `eglDestroyDisplayEXT` and `eglTerminate` is that it doesn't allow reviving the display? If so, that's exactly what I wanted I think. Thanks!

kbrenneman:
> Cool, sounds good. Just checking, the difference between `eglDestroyDisplayEXT` and `eglTerminate` is that it doesn't allow reviving the display? If so, that's exactly what I wanted I think. Thanks!

In effect, yeah. Every EGL function would do exactly the same thing that it does now. `eglTerminate` basically reverts the EGLDisplay to the same uninitialized state that it was in before you called `eglInitialize`. Trying to pass it to any other function would produce an `EGL_NOT_INITIALIZED` error, but the EGLDisplay handle still exists, so you can pass it to `eglInitialize`. If you call `eglGetPlatformDisplay` with `EGL_PRIVATE_DISPLAY_EXT` set to true, then it will *not* return that EGLDisplay handle.

When you call `eglDestroyDisplayEXT`, it destroys the EGLDisplay handle entirely, with the same semantics you'd expect out of `eglDestroyContext`. You can't reinitialize it because it no longer exists, and a call to `eglGetPlatformDisplay` is free to re-use that handle.

> Edit: out of curiosity, what codebase are you implementing this on? I was hoping to implement it on my fork of ANGLE.

Libglvnd first, since that'll be needed for anything else to work. After that, the NVIDIA driver.

nwnk:
I've started sketching this up for [Mesa](https://gitlab.freedesktop.org/ajax/mesa/-/commit/ecebf3a0074877367412544688f64a2ddc496ab8) and it's looking pretty sane. I might have some questions about exactly what gets destroyed when but I think the discussion above has most of the answers. I'll link to a merge request once I've got it working but I think the spec gets a +1 from me.

nwnk:
[Mesa merge review](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/17604) is now up. @kbrenneman if you have any tests for this you could share that'd be cool. (I do not, yet, they'll be in piglit once I do though.)

emersion:
Any news about this ext?

kchibisov:
What should happen if I create 2 displays (`A` and `B`) using the same platform display (e.g. `wl_display'). And then use `EGLSurface` from display `A` in a `EGLContext` from display `B`.

Do I understand correctly, that `unshared` means that I can't use anything from `A` with the `B` entirely even though, they have exactly the same platform display under the hood?

emersion:
Each EGLDisplay would have its own `wl_display`, and Wayland objects cannot be shared across two `wl_display`s.

kchibisov:
>Each EGLDisplay would have its own wl_display, and Wayland objects cannot be shared across two wl_displays.

I don't think I understand with `wl_display`(you don't `wl_display_connect` you just get the handle from the user), but I think you've meant that they will have different queues under the hood and you can't mix the objects between the queues.

However on X11 where everything is a XID, I think you could be able to do so, given that.

>    If the EGL_PRIVATE_DISPLAY_EXT attribute is set to EGL_TRUE, then
>    eglGetPlatformDisplay returns a new EGLDisplay handle, even if the
>    platform, native_display, and attributes would match an existing
>    display.

It would be nice to be more explicit here, so the libraries could do runtime asserts just in case.

emersion:
Oh, hm, right. You should be able to use `wl_display` objects from multiple queues… Assuming the per-queue wrappers are properly created.

kbrenneman:
You can create two EGLDisplays from the same wl_display, but they're still separate EGLDisplays.

Remember, an EGLDisplay is the namespace that EGLContexts, EGLSurfaces, and other EGL objects belong to. Every other EGL object is *only* valid with the EGLDisplay that created it. So, if you try to call eglMakeCurrent with EGLDisplay A, but you pass it an EGLContext from EGLDisplay B, then that's an error.

That much is not new behavior. You can already create multiple EGLDisplays from the same native display just by passing different attributes to eglGetPlatformDisplay. I think Mesa will hand back different EGLDisplays even if the attribute lists are logically equivalent:
```C
wl_display *wdpy = wl_display_connect(NULL);
EGLAttrib attribs1[] = [ EGL_TRACK_REFERENCES_KHR, EGL_TRUE, EGL_NONE ]
EGLDisplay dpy1 = eglGetPlatformDisplay(EGL_PLATFORM_WAYLAND_KHR, wdpy, attribs1);
EGLAttrib attribs2[] = [ EGL_TRACK_REFERENCES_KHR, EGL_FALSE, EGL_NONE ]
EGLDisplay dpy2 = eglGetPlatformDisplay(EGL_PLATFORM_WAYLAND_KHR, wdpy, attribs2);
EGLAttrib attribs3[] = [ EGL_TRACK_REFERENCES_KHR, EGL_FALSE, EGL_NONE ]
EGLDisplay dpy3 = eglGetPlatformDisplay(EGL_PLATFORM_WAYLAND_KHR, wdpy, attribs3);
EGLAttrib attribs4[] = [ EGL_NONE ]
EGLDisplay dpy4 = eglGetPlatformDisplay(EGL_PLATFORM_WAYLAND_KHR, wdpy, attribs4);
if (dpy1 != dpy2) {
    // This will be true on all drivers, because we used different attribute lists.
}
if (dpy2 == dpy3) {
    // This will be true on all drivers, because we used the same attribute list.
}
if (dpy3 != dpy4) {
    // This could be true or false depending on the driver.
    // EGL_TRACK_REFERENCES_KHR defaults to false, so the attribute lists are
    // different, but logically equivalent. The EGL spec allows either behavior.
}
EGLContext ctx = eglCreateContext(dpy1, ...);
eglMakeCurrent(dpy2, NULL, NULL, ctx); // ERROR: ctx came from dpy1, not dpy2
```

What this extension adds is a way to *guarantee* that you get a different EGLDisplay when you call eglGetPlatformDisplay, even if the attribute list is otherwise identical. It also gives you a way to destroy an EGLDisplay afterward instead of leaking it.

stonesthrow:
When you say "private" EGL_PRIVATE_DISPLAY_EXT, private to what? Is this private to a single thread (TLS space for implementation)? or private by some other context? That would need to be clear in the spec.

kbrenneman:
"Private" in this case means private to whatever called eglGetPlatformDisplay, rather than the default behavior where everything that calls eglGetPlatformDisplay would end up with the same shared EGLDisplay object.

The use I was thinking about initially was for libraries that use EGL as an internal implementation detail, where they'd want their own private EGLDisplay so that they don't interfere with the application or with other libraries.

The name "EGL_PRIVATE_DISPLAY" is left over from that, but it's definitely not the best choice of enum name.

Maybe something like "EGL_ALLOC_NEW_DISPLAY_EXT" instead? That would more clearly describe what it does, rather than what it might be used for.

I'm open to other suggestions, if anyone's got any.

cubanismo:
Is "UNIQUE" any clearer?

stonesthrow:
I just wanted to be sure that it wasn't private to a thread. 
eglGetPlatformDisplay" with EGL_PRIVATE_DISPLAY_EXT set EGL_TRUE - will not return the same EGLDisplay handle twice, so private by unknown/obscure by any other EGL user.

kbrenneman:
Yeah, that's correct.

stonesthrow:
Looking at your current EGL_EXT_display_alloc.txt
lines 62-64 Good
Line 88, may I suggest add "unique" handle. That might cover James' suggestion and would cover my desire for clarity for users and implementers.

kbrenneman:
Can do.

For the enum name, would something like "EGL_UNIQUE_DISPLAY_EXT" or "EGL_ALLOC_DISPLAY_EXT" read better than "EGL_PRIVATE_DISPLAY_EXT"?

stonesthrow:
PRIVATE is OK with me. In all cases what that means needs to be clear.

stonesthrow:
Are suspicious of any issue with multiple EGLDisplay handles referring to one Window System. Since this is an EXT, they can bow out, but just curious if it could be a problem anywhere? Sorry if this was discussed before

stonesthrow:
Needs rebase. Add any changes, and any issues/resolution from this discussion you think informative.
I'll give a +1 to the review

kbrenneman:
> Are we suspicious of any issue with multiple EGLDisplay handles referring to one Window System. Since this is an EXT, they can bow out, but just curious if it could be a problem anywhere? Sorry if this was discussed before

It wouldn't be any more of a problem with this extension than it was without. Any EGL implementation already has to deal with calls to eglGetPlatformDisplay with the same native display and different attribute lists.

kchibisov:
>Is "UNIQUE" any clearer?

I find `UNIQUE` a bit of confusing myself(not that my opinion really matters though), given that it works with `EGL_KHR_display_reference` extension. I'd suggest `OWNED`(or a variation of this word), because it's a common word in `reference`/`lifetimes` contexts and that is what we get - an `OWNED` display behind a `pointer`/`reference counted container`, where the default(as in default display semantic) is a shared proxy to a singleton-like object.

stonesthrow:
> > Are we suspicious of any issue with multiple EGLDisplay handles referring to one Window System. Since this is an EXT, they can bow out, but just curious if it could be a problem anywhere? Sorry if this was discussed before
> 
> It wouldn't be any more of a problem with this extension than it was without. Any EGL implementation already has to deal with calls to eglGetPlatformDisplay with the same native display and different attribute lists.

Agreed. Good

kbrenneman:
> > Is "UNIQUE" any clearer?
> 
> I find `UNIQUE` a bit of confusing myself(not that my opinion really matters though), given that it works with `EGL_KHR_display_reference` extension. I'd suggest `OWNED`(or a variation of this word), because it's a common word in `reference`/`lifetimes` contexts and that is what we get - an `OWNED` display behind a `pointer`/`reference counted container`, where the default(as in default display semantic) is a shared proxy to a singleton-like object.

This extension is entirely orthogonal to EGL_KHR_display_reference. You can use them together, or you can use either one by itself.

That was similar to my reasoning for something like `ALLOC_DISPLAY` or `ALLOC_NEW_DISPLAY` for a name. That does seem clear about what it does, but it feels a little weird to use an imperative for an attribute name.

kchibisov:
>This extension is entirely orthogonal to EGL_KHR_display_reference. You can use them together, or you can use either one by itself.

I understand that, it's just for me if you use them together `unique` is a bit weird, given that you'll have a *reference to unique* display, where what you have is *reference to your owned* display, given that default(without this extension) display is borrowed (i.e. you have a reference to global data).

cubanismo:
Naming things is always hard. If UNIQUE sounds too weird with a reference usage, I think I'm back to PRIVATE as being the ideal name. Private extensions are never shared implicitly, only explicitly, as opposed to the default (Public), which are always shared unless someone creates a new Private one. That flows more naturally than "Owned" or "Borrowed" to me. Borrowing implies a temporary transfer of ownership, and that isn't the correct description for use of a display without this extension either. They're shared, but that could also apply to "Owned" displays that are referenced, so it isn't a good antonym for "Owned" either.

stonesthrow:
"Unique" is OK for me. Because in the text I pointed out - caller gets a unique handle. Meaning caller holds the only copy and responsible for the terminate and destroy. I guess it was the right word for me to grasp. But its your extension.

api/egl.xml
```diff
@@ -817,7 +817,7 @@
         <enum value="0x3328" name="EGL_PENDING_METADATA_NV"/>
         <enum value="0x3329" name="EGL_PENDING_FRAME_NV"/>
         <enum value="0x332A" name="EGL_STREAM_TIME_PENDING_NV"/>
-            <unused start="0x332B"/>
+        <enum value="0x332B" name="EGL_PRIVATE_DISPLAY_EXT"/>
```
stonesthrow, line 820: Did you already check with James to claim this token enum?

stonesthrow (commented):
Otherwise OK +1

api/egl.xml
kbrenneman: Yeah, I checked before I posted my initial draft.

stonesthrow (approved):
LGTM

stonesthrow:
Any more reviewers/comments?

oddhack:
@stonesthrow @kbrenneman @cubanismo ISTM this is in a publishable state (aside from needing a rebase and change to the extension # as that's out of date) - I don't see any feedback in the comments that does not seem to have been addressed. Could this be moved to Signed Off milestone?

kbrenneman:
The only thing I know of that's still open is the enum name. `EGL_PRIVATE_DISPLAY_EXT` is what I've still got in the spec, which isn't as clear a name as it ought to be, and carries some incorrect implications.

`EGL_UNIQUE_DISPLAY_EXT` and `EGL_ALLOC_NEW_DISPLAY_EXT` have come up as alternatives.

Personally, I'd lean slightly toward `EGL_ALLOC_NEW_DISPLAY_EXT`, since that's a pretty direct description of what it actually does.

stonesthrow:
Do want to change the enum name? then we'll approve if ready.

oddhack:
> Do want to change the enum name? then we'll approve if ready.

Not my call but agreed that EGL_ALLOC_NEW_DISPLAY_EXT is best of the alternatives.

kbrenneman:
I'll change it to `EGL_ALLOC_NEW_DISPLAY_EXT`, then.

stonesthrow (approved):
LGTM

stonesthrow:
0x332B  was reserved for James. If claiming use of this enum is OK, let me know and I will move this to "Approved for merge".

oddhack:
> 0x332B was reserved for James. If claiming use of this enum is OK, let me know and I will move this to "Approved for merge".

That's a question for @cubanismo, so tagged them in as a reviewer.

oddhack (approved):
LGTM, pending @cubanismo signoff on use of the NV reserved enum value for this EXT.

cubanismo:
It's been a long time since I did any EGL extension work. Was it reserved specifically for me, or for NV? If the latter, Kyle's at NV too, so that's fine assuming he used our internal tracking stuff to assign it to this extension.

oddhack:
> It's been a long time since I did any EGL extension work. Was it reserved specifically for me, or for NV?

I'd have to do a lot of archival work since it was a Bugzilla commit and we've been on git for many years. Generally someone from the vendor reserves the space and what happens to it after that is an internal matter. If you have internal tracking to avoid collisions then there's probably no issue.

cubanismo:
OK, thanks. We'll resolve internally. Just thought you might be referencing some other info I couldn't find.

stonesthrow (commented):
egl.xml line 815 - block reserved for James.
I just wanted to be sure that using the open enum for this extension was OK with you James or NV.

kbrenneman:
I'll double-check our internal enum bookkeeping.

kbrenneman:
I updated the enum based on our internal bookkeeping.

The spec should be ready to publish, unless we want to wait until there's an actual implementation ready to go. I'm not sure when that'll be for the NVIDIA driver -- the external platform library stuff would make that more complicated than it would be for Mesa.

api/egl.xml
```diff
@@ -886,7 +886,8 @@
         <enum value="0x3376" name="EGL_STREAM_IMAGE_AVAILABLE_NV"/>
         <enum value="0x3377" name="EGL_DRM_RENDER_NODE_FILE_EXT"/>
         <enum value="0x3378" name="EGL_STREAM_CONSUMER_IMAGE_USE_SCANOUT_NV" />
-            <unused start="0x3379" end="0x339F"/>
+        <enum value="0x337A" name="EGL_ALLOC_NEW_DISPLAY_EXT"/>
```
stonesthrow, line 889: 337A is your unused, 3379 is what you have in the extension spec below.

stonesthrow (changes requested):
Fix typo

api/egl.xml
kbrenneman: Agh, sorry about that. Fixed.

stonesthrow (approved):

stonesthrow:
Had to look up the Extension rules:

Multi-Vendor Extensions. Essentially Vendor Extensions developed and supported by multiple organizations, with no collaboration or permissions needed from Khronos (but optionally using collaboration within the working group). **A Multi-Vendor Extension is not ratified by Khronos and so has no implementations or conformance requirements** for release but should be entered into the registry through requesting registry enumerants from the registry maintainer. A Multi-Vendor Extension should typically use the ‘EXT’ prefix.

oddhack (approved):

PR #160, opened 2022-07-07
EGL_KHR_platform_wayland: clarify surface querying

kchibisov:
When querying EGLSurface on Wayland platform it'll result in locking
its back buffer until the next swap. This behavior is long established
in mesa and is expected.

The rationale behind it is to keep data returned from the
eglQuerySurface and rendering operations applied to that surface
accurate until the next swap.

This commit explicitly mentions that behavior.

--

This is a draft for now, since I'm not sure how to word it and  submit patches to specs. It's also a MESA specific issue, but there's nvidia as well? What about `EGL_EXT_platform_wayland`?

Also, should we mention that calling eglCreatePlatformPbufferSurface will also always fail the way pixmaps do in a follow-up patch? 

cc @fooishbar

### DIFF
--- a/extensions/KHR/EGL_KHR_platform_wayland.txt
+++ b/extensions/KHR/EGL_KHR_platform_wayland.txt
@@ -40,6 +40,9 @@ Dependencies
     This extension is written against the EGL 1.5 Specification (draft
     20140122).
 
+    The behavior of part of this extension is different depending on whether
+    the EGL_EXT_buffer_age extension is also present.
+
 Overview
 
     This extension defines how to create EGL resources from native Wayland
@@ -85,6 +88,11 @@ New Behavior
     belongs to Wayland. Any such call fails and generates an
     EGL_BAD_PARAMETER error.
 
+    Rendering to the obtained EGLSurface or querying it with EGL_BUFFER_AGE_KHR
+    will lock its back buffer preventing it from being dropped or resized,
+    until the next buffer swap. The rationale behind this behavior is to keep
+    operations result accurate until the next swap.
+
 Issues
 
     1. Should this extension permit EGL_DEFAULT_DISPLAY as input to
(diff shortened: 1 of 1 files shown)


kchibisov:
>Such as switch to EGL_SINGLE_BUFFER which would be a change to the buffer count

Wayland with mesa always uses double buffering. It doesn't support a way to be single buffered, either via creation or changing (not ext for changing is implemented).

>Add text telling impact of surface lock on eglSurfaceAttrib; are there new error cases? 

It doesn't error out from what I know, but you can't do much with it on Wayland with mesa, except to change EGL_MIPMAP_LEVEL, which returns true after quering buffer age.

I'm not an expert in mesa internals though.

So from what I can see in the particular implementation it doesn't affect the `eglSurfaceAttrib`, but I'd like to wait for @fooishbar reply as well on that matter.

stonesthrow:
> Wayland with mesa always uses double buffering.

OK. if so, then this is not a scenario to worry about.

extensions/KHR/EGL_KHR_platform_wayland.txt
```diff
@@ -85,6 +85,11 @@ New Behavior
     belongs to Wayland. Any such call fails and generates an
     EGL_BAD_PARAMETER error.
 
+    Rendering to obtained EGLSurface or querying it with eglQuerySurface will
```
fooishbar, line 88: Hmmmm, I think we should explicitly call out `EGL_BUFFER_AGE_KHR` as that's the only query I can think of. Querying, e.g., `EGL_WIDTH` and `EGL_HEIGHT` doesn't lock in a back buffer.

kchibisov: I've updated wording.

stonesthrow (approved):
Reads clear and helpful to users.

fooishbar:
Thanks, this looks great! Do we also need to reference `EGL_EXT_buffer_age` in the dependencies, or is that OK?

@kchibisov you should also add an entry to the changelog at the bottom with your name and the change

kchibisov:
>@kchibisov you should also add an entry to the changelog at the bottom with your name and the change

Yeah, I wasn't sure about this, so that's why it's a draft.

stonesthrow:
"Do we also need to reference EGL_EXT_buffer_age in the dependencies, or is that OK?"
That would be great and correct

kchibisov:
I've added a dependency, updated revision history and commit message.

fooishbar:
Awesome, looks great to me, approved

oddhack:
> Approved with reviews by Daniel and I. OddHack will check it over and merge the change

LGTM.

Issue #166, opened 2022-08-10
eglCreateContext attrib_list documentation is confusing


vvavrychuk:
https://registry.khronos.org/EGL/sdk/docs/man/html/eglCreateContext.xhtml attrib_list parameter explanation says

> Only the attribute EGL_CONTEXT_CLIENT_VERSION may be specified.

Word `may` looks like a typo, since in `eglCreateContext` function description a lot more attributes are mentioned which **may** be specified.

stonesthrow:
Agreed this sentence  is funny given the list of allowed attributes.

I suggest we remove this "Only the attribute EGL_CONTEXT_CLIENT_VERSION may be specified."

stonesthrow:
Thanks @vvavrychuk . I will make a change and have it reviewed per EGL process.

vvavrychuk:
Thanks @stonesthrow for quick feedback!

PR #167, opened 2022-08-10
Update eglCreateContext.xml

stonesthrow:
Closes: https://github.com/KhronosGroup/EGL-Registry/issues/166

Remove: " Only the attribute <constant>EGL_CONTEXT_CLIENT_VERSION</constant> may be specified."

### DIFF
--- a/sdk/docs/man/eglCreateContext.xml
+++ b/sdk/docs/man/eglCreateContext.xml
@@ -62,9 +62,7 @@
                 <term><parameter>attrib_list</parameter></term>
                 <listitem><para>
                     Specifies attributes and attribute values for the
-                    context being created. Only the attribute
-                    <constant>EGL_CONTEXT_CLIENT_VERSION</constant> may
-                    be specified.
+                    context being created.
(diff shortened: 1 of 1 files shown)


stonesthrow:
Note that eglCreateContext has a number of attributes, so this comment isn't valid.
https://registry.khronos.org/EGL/sdk/docs/man/html/eglCreateContext.xhtml

vvavrychuk:
Just to double check.

I wander if original intention of this sentence was
```
Only the attribute <constant>EGL_CONTEXT_CLIENT_VERSION</constant> **must** be specified.
```

Just guessing...

But, on the other hand it contradicts with EGL_CONTEXT_MAJOR_VERSION documented having default version 1. So probably this assumption is wrong.

oddhack (approved):
I don't recall the origins of that language at a couple of decades remove, but the change is fine.

vvavrychuk:
LGTM

stonesthrow:
Using you as third reviewer Jon. Small change, approving.

Issue #169, opened 2022-08-24
eglCreatePlatform{Window,Pixmap}Surface documentation should mention EGL 1.5 dependency


RAOF:
As far as I can tell these entry points are new in EGL 1.5 (folded in from the `EGL_EXT_platform_base` extension in 1.4). The `eglGetPlatformDisplay` correctly mention that it's new in 1.5, but neither the `eglCreatePlatformWindowSurface` nor `eglCreatePlatformPixmapSurface` docs do (instead they somewhat-misleadingly document some of their accepted parameter values as new in 1.5).

I can prepare a PR to update the docs if I've understood this correctly.

stonesthrow:
As I recall, yes, they were extension in EGL 1.4 and then with EGL 1.5 they were accepted as core API.
Yes, Please point out the inconsistent/misleading text for correction. thx.

stonesthrow:
Of course, the names of functions and tokens will change - dropping "EXT" or "KHR" when adopted into core.

PR #170, opened 2022-08-24
Clarify version requirements of CreatePlatform{Window,Pixmap}Surface

RAOF:
Fixes: #169

These were folded into EGL 1.5 from EGL_EXT_platform_base. Make that clear, and remove
notes about parameter values newly available in EGL 1.4 and EGL 1.2

Signed-off-by: Christopher James Halse Rogers <raof@ubuntu.com>

### DIFF
--- a/sdk/docs/man/eglCreatePlatformPixmapSurface.xml
+++ b/sdk/docs/man/eglCreatePlatformPixmapSurface.xml
@@ -164,7 +164,7 @@
     </refsect1>
     <refsect1 xml:id="notes"><title>Notes</title>
         <para>
-            <code class="function">eglCreatnePlatformPixmapSurface</code> is supported only if
+            <code class="function">eglCreatePlatformPixmapSurface</code> is supported only if
             the EGL version is 1.5 or greater.
         </para>
         <para>
--- a/sdk/docs/man/eglCreatePlatformPixmapSurface.xml
+++ b/sdk/docs/man/eglCreatePlatformPixmapSurface.xml
@@ -164,22 +164,15 @@
     </refsect1>
     <refsect1 xml:id="notes"><title>Notes</title>
         <para>
+            <code class="function">eglCreatnePlatformPixmapSurface</code> is supported only if
+            the EGL version is 1.5 or greater.
+        </para>
+        <para>
             The <constant>EGL_MATCH_NATIVE_PIXMAP</constant> attribute
             of
             <citerefentry><refentrytitle>eglChooseConfig</refentrytitle></citerefentry>
             can be used to select a frame buffer configuration matching
             a specified native pixmap.
-        </para>
-        <para>
-            Attribute <constant>EGL_GL_COLORSPACE</constant> is supported
-            only if the EGL version is 1.5 or greater.
-        </para>
-        <para>
-            Attributes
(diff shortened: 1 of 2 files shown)


RAOF:
I've not regenerated the HTML docs, as that introduced a bunch of spurious diff. I can easily do that if it's helpful though.

oddhack:
> I've not regenerated the HTML docs, as that introduced a bunch of spurious diff. I can easily do that if it's helpful though.

I'll do it after the PR is merged.

We don't tag all the refpages with the corresponding EGL version but no harm in doing so here. LGTM.

RAOF:
Thanks for the typo fix :grimacing:

stonesthrow:
+ @cubanismo for review, if you have a moment. thx.

sdk/docs/man/eglCreatePlatformPixmapSurface.xml
```diff
@@ -163,24 +163,17 @@
...
+            the EGL version is 1.5 or greater.
+        </para>
         <para>
             The <constant>EGL_MATCH_NATIVE_PIXMAP</constant> attribute
             of
             <citerefentry><refentrytitle>eglChooseConfig</refentrytitle></citerefentry>
             can be used to select a frame buffer configuration matching
             a specified native pixmap.
         </para>
-        <para>
```
stonesthrow, line 173: From the spec appendix and egl.h versioning. History looks to be. COLORSPACE was added with EGL 1.2. With EGL 1.3 the VG_COLORSPACE was added. Then in EGL 1.5 GL_COLORSPACE was added to core.
GL_COLORSPACE was/is an extension prior to EGL 1.5.
So, maybe not delete but modify these notes? Jon, what do you recommend?

stonesthrow (commented):
Remove or modify notes? - @oddhack what do you think?

sdk/docs/man/eglCreatePlatformPixmapSurface.xml
cubanismo: Agreed. I'm not clear why these notes should be removed. It would be nice to clarify them.

stonesthrow: So for first part: only if EGL 1.5 or extension EGL_KHR_gl_colorspace is supported.

Second part could stay.

stonesthrow (commented):
Suggested change

sdk/docs/man/eglCreatePlatformPixmapSurface.xml
RAOF: Hm. I removed the notes because I didn't think they're relevant to users of `eglCreatePlatform{Pixmap,Window}Surface` - the entry point only exists in EGL 1.5, users are guaranteed to be able to use those attributes, since they're all core in 1.5?

I'm happy to add them back if that's desired, though.

stonesthrow: OK. Yes. Since this is man pages, it doesn't take in to account extensions. So you are right, its an entry point only dependent on EGL 1.5.

stonesthrow (approved):
OK this is sufficient. thanks

PR #182, opened 2023-07-01
EGL_EXT_yuv_surface: replace symbol "ﬁ" with letters "f" and "i" in the word "defined"

1ace:


stonesthrow:
Interesting typo. Thanks for fix.

PR #185, opened 2023-10-30
EGL_EXT_create_context_robustness: query reset notification strategy through eglQueryContext

### LINKED ISSUE #184: Query context reset notification strategy
With the extension `EGL_EXT_create_context_robustness` [1], it is possible to specify the reset notification strategy of a context by adding two fields in the attribute list when using `eglCreateContext`.

Unfortunately, it is not possible to query the reset notification strategy of a given context. When trying to get the `EGL_CONTEXT_OPENGL_RESET_NOTIFICATION_STRATEGY_EXT` with `eglQueryContext`, `EGL_BAD_ATTRIBUTE` is returned, despite the requested attribute being valid.

This comes in the way when trying to create a shared context of a context with a specific reset notification strategy. From the `EGL_EXT_create_context_robustness` spec changes:

      If the reset notification behavior of <share_context> and the
      newly created context are different then an EGL_BAD_MATCH error is
      generated.

Is this an implementation details, and Mesa (the driver I am using) could returns the correct value instead of `EGL_BAD_ATTRIBUTE`, or is this a whole in the spec waiting to be defined and improved?

[1]: https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_create_context_robustness.txt

bl4ckb0ne:
Closes #184 

Mesa implementation [1] and usage in Monado [2]

[1]: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25961
[2]: https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2004/diffs?commit_id=43e7543d6276841e1bf21c91e07978235a15645b

### DIFF
--- a/extensions/EXT/EGL_EXT_create_context_robustness.txt
+++ b/extensions/EXT/EGL_EXT_create_context_robustness.txt
@@ -21,7 +21,7 @@ Status
 
 Version
 
-    Version 3, 2011/10/31
+    Version 4, 2023/10/30
 
 Number
 
@@ -114,7 +114,7 @@ Additions to the EGL 1.4 Specification
     upon reset notification as described by the GL_EXT_robustness. The
     default value for EGL_CONTEXT_OPENGL_RESET_NOTIFICATION_STRATEGY_EXT
     is EGL_NO_RESET_NOTIFICATION_EXT.
-    
+
     Add to the eglCreateContext context creation errors:
 
     * If <config> does not support a client API context compatible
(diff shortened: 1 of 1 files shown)


oddhack:
This changes the behavior of a shipping extension. What will an app do that expects this behavior when running against an implementation that has not been updated accordingly?

@dgkoch you might care to comment since this was your extension originally, albeit in a previous life?

dgkoch:
👎 Yeah I don't think adding functionality to an extension that has been shipping for 12 years is a good idea.

bl4ckb0ne:
I see your point, I'll close it.

PR #186, opened 2023-11-01
introduce EGL_EXT_query_reset_notification_strategy

### LINKED ISSUE #184: Query context reset notification strategy

bl4ckb0ne:
Closes #184 

Mesa implementation [1] and usage in Monado [2]

[1]: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25996
[2]: https://gitlab.freedesktop.org/monado/monado/-/merge_requests/2004

### DIFF
--- a/extensions/EXT/EGL_EXT_query_reset_notification_strategy.txt
+++ b/extensions/EXT/EGL_EXT_query_reset_notification_strategy.txt
@@ -1,0 +1,79 @@
+Name
+
+	EXT_query_reset_notification_strategy
+
+Name Strings
+
+	EGL_EXT_query_reset_notification_strategy
+
+Contributors
+
+	Simon Zeni <simon.zeni@collabora.com>
+
+Contact
+
+	Simon Zeni <simon.zeni@collabora.com>
+
+Status
+
+	Draft
+
+Version
+
+	Version 0.1, 2023-11-01
+
+Number
+
+	EGL Extension #152
+
+Extension Type
+
+	Client
+
+Dependencies
+
+	Requires EGL 1.4 and EXT_create_context_robustness.
+
+	The extension is written against the wording of the EGL 1.5
+	specification, but may be implemented against 1.4.
+
+Overview
+
+	This extension complements EXT_create_context_robustness and enables
+	an application or framework to retrieve an existing context's reset
+	notification strategy in order to create a compatible shared context.
+
+New Types
+
+	None
+
+New Procedures and Functions
+
+	None
+
+New Tokens
+
+	None
+
+Specification modifications
+
+	Append in section 3.7.4 (Context Queries), to the list of attributes
+	supported by eglQueryContext.
+
+		Querying EGL_CONTEXT_OPENGL_RESET_NOTIFICATION_STRATEGY_EXT
+		returns the reset notification strategy that was set for this
+		context. If no strategy was set, EGL_NO_RESET_NOTIFICATION_EXT
+		must be returned.
+
+Errors
+
+	None
+
+Issues
+
+	None
+
+Revision History
+
+	Version 0.1, 2023-11-01 (Simon Zeni)
+		- Initial draft
--- a/registry.tcl
+++ b/registry.tcl
@@ -783,4 +783,9 @@ extension EGL_QNX_image_native_buffer {
     flags       public
     filename    extensions/QNX/EGL_QNX_image_native_buffer.txt
 }
-# Next free extension number: 152
+extension EXT_query_reset_notification_strategy {
+    number      152
+    flags       public
+    filename    extensions/EXT/EGL_EXT_query_reset_notification_strategy.txt
+}
+# Next free extension number: 153
--- a/api/egl.xml
+++ b/api/egl.xml
@@ -2685,6 +2685,7 @@
                 <enum name="EGL_PROTECTED_CONTENT_EXT"/>
             </require>
         </extension>
+        <extension name="EGL_EXT_query_reset_notification_strategy" supported="egl"/>
         <extension name="EGL_EXT_stream_consumer_egloutput" supported="egl">
             <require>
                 <command name="eglStreamConsumerOutputEXT"/>


stonesthrow (approved):
Looks complete to me

oddhack (changes requested):
This should still get an empty interface in egl.xml (see e.g. EGL_ANDROID_GLES_layers) but otherwise is fine. Note that the extension tag should go in the right place in the overall extension tag sort order.

bl4ckb0ne:
Added the entry to `egl.xml`.

>Note that the extension tag should go in the right place in the overall extension tag sort order.

What do you mean? In the `egl.xml` file?

extensions/EXT/EGL_EXT_query_reset_notification_strategy.txt
```diff
@@ -0,0 +1,78 @@
...
+
+	Requiers EGL 1.4 and EXT_create_context_robustness.
+
+	The extension is written against the wording of the EGL 1.5
+	specification, but may be implemented against 1.4.
+
+Overview
+
+	This extension allows the user to query the reset notification strategy
+	of a given context using the function eglQueryContext.
```
dgkoch, line 43: Maybe you could add a bit more motivation to the overview for why this is needed/what problem it solves?

extensions/EXT/EGL_EXT_query_reset_notification_strategy.txt
```diff
@@ -0,0 +1,78 @@
...
+
+Specification modifications
+
+	Append in section 3.7.4 (Context Queries), to the list of attributes
+	supported by eglQueryContext.
+
+		Querying EGL_CONTEXT_OPENGL_RESET_NOTIFICATION_STRATEGY_EXT
+		returns the reset notification strategy that was set for this
+		context. If no strategy was set, EGL_NO_RESET_NOTIFICATION_EXT
+		must be returned.
```
dgkoch, line 65: What if it's not a robust context at all?  Is that intended to be the same as "no strategy set"?

dgkoch (changes requested):

extensions/EXT/EGL_EXT_query_reset_notification_strategy.txt
bl4ckb0ne: Mesa defaults it to `EGL_NO_RESET_NOTIFICATION_KHR` regardless of a robust context or not[1]

[1]: https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/egl/main/eglcontext.c?ref_type=heads#L615

dgkoch: Probably, although this extension says it requires EXT_create_context_robustness, so I'm not sure how you'd get into that situation..

extensions/EXT/EGL_EXT_query_reset_notification_strategy.txt
```diff
@@ -0,0 +1,80 @@
...
+	Requires EGL 1.4 and EXT_create_context_robustness.
+
+	The extension is written against the wording of the EGL 1.5
+	specification, but may be implemented against 1.4.
+
+Overview
+
+	This extension allows the user to query the reset notification strategy
+	of a given context using the function eglQueryContext. This extension
+	complements the EXT_create_context_robustness one in order to be able
```
dgkoch, line 44:
```suggestion
	complements EXT_create_context_robustness in order to be able
```

This is still a bit obtuse, since this extension has nothing to do with creating contexts.

Maybe something like: 

This extension complements EXT_create_context_robustness and enables an application or framework to determine an existing context's reset notification strategy in order to enable it to create a compatible shared context.

oddhack:
> Added the entry to `egl.xml`.
> 
> > Note that the extension tag should go in the right place in the overall extension tag sort order.
> 
> What do you mean? In the `egl.xml` file?

Right. They are in order by name.

bl4ckb0ne:
Oh I see. I got confused because some recent extensions are just added at the bottom of the file.

extensions/EXT/EGL_EXT_query_reset_notification_strategy.txt
bl4ckb0ne: I reworded it based on your feedback, thanks

extensions/EXT/EGL_EXT_query_reset_notification_strategy.txt
bl4ckb0ne: With the requirement, I think it's implicit that the query returns `EGL_BAD_ATTRIBUTE` if `EXT_create_context_robustness` is not supported. 

Should it still return `EGL_NO_RESET_NOTIFICATION_KHR` if the context is not robust? Or should we introduce a new value to be returned? I quickly looked at `GL_EXT_robustness` and I don't see a different behaviour if the context is not robust.

dgkoch: I guess it's fine then. I haven't thought about EGL much in a few years now! ;-)

dgkoch (approved):
Looks fine now.

stonesthrow:
Requested changes need resolved/closed. I will then approve, and Jon can review and merge.

bl4ckb0ne:
Ping @oddhack for the last review on the requested changes

registry.tcl
```diff
@@ -783,4 +783,9 @@ extension EGL_QNX_image_native_buffer {
     flags       public
     filename    extensions/QNX/EGL_QNX_image_native_buffer.txt
 }
-# Next free extension number: 152
+extension EGL_EXT_query_context {
```
stonesthrow, line 786: extension name is not a match

stonesthrow (changes requested):
copy-paste typo?

registry.tcl
bl4ckb0ne: I renamed the extension halfway through. Thanks for catching this.

stonesthrow (approved):

stonesthrow:
Approving from Daniel and My review. Turn over to Jon for review and merge.

oddhack (approved):
LGTM.

PR #187, opened 2023-11-07
Update eglQueryAPI.xml

rpavlik:
Fix an apparent typo: this is the only revision that made sense to me.


stonesthrow (commented):
Agreed, the first part of the note matches section 3.7 of spec. The "however" clause makes sense in that its invalid to use with eglBindAPI, and this is consistent with the rest of the text here under "eglQueryAPI".

stonesthrow:
@oddhack , please check this typo.

oddhack:
LGTM. You can set Approved to Merge or just merge it (I'll have to regenerate the HTML refpages though, so probably best if I merge it).

oddhack:
Page regenerated.

PR #190, opened 2023-11-22
Create SECURITY.md

tylern12:


oddhack:
This project does not have "versions" or "releases". It is backing store for the corresponding web pages, and a few scripts used internally to update parts of the store.

PR #198, opened 2024-02-28
add fix for lxml import based on suggestion from lxml.de/tutorial.html

pcm1289:
After installing the `lxml` dependency in MSYS on Windows using:
```
pacman -S python-lxml
```

..there was still an error:
```
$ make
```
```
python3 -B genheaders.py  -registry egl.xml EGL/egl.h
Traceback (most recent call last):
  File "EGL-Registry/api/genheaders.py", line 7, in <module>
    from reg import *
  File "EGL-Registry/api/reg.py", line 7, in <module>
    from lxml import etree
ImportError: No such process
make: *** [Makefile:17: EGL/egl.h] Error 1
```
image

It appears as if these may be well known because this website presented an alternative for those not depending explicitly on functions included in the `lxml` implementation of `etree`, found here:
https://lxml.de/tutorial.html

Upon making this change..
image

..the ability to use `make` out of the box was restored:
image

This pull request addresses this. A visual inspection of `reg.py` on surface level looks to not use any `lxml.etree` specific functions.

### DIFF
--- a/api/reg.py
+++ b/api/reg.py
@@ -4,7 +4,18 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import io,os,re,string,sys
-from lxml import etree
+try:
+    # According to https://lxml.de/tutorial.html
+    #   if we are not using any lxml etree specific functions,
+    #   then xml.etree.ElementTree should behave the same way.
+    # Due to differences using MSYS on Windows, using xml.etree
+    #   instead was tested and seems fine.
+    from lxml import etree
+    print("running with lxml.etree")
+except ImportError:
+    import xml.etree.ElementTree as etree
+    print("running with Python's xml.etree.ElementTree")
+
 import subprocess
 
 def write(*args, **kwargs):


stonesthrow:
@oddhack can you review this?

oddhack:
> > @oddhack can you review this?
> 
> Sorry, I overlooked this. It is true that ElementTree is API-compatible for the functionality we use here, but it actually isn't semantically compatible in some corner cases including one we might use in the scripts. I'll take a look and if the outputs are identical, we can just use ElementTree.

Verified we can replace lxml entirely for purposes of this repo. Suggest closing this and I'll merge #200 instead.

oddhack:
@pcm1289 could you look at #200 and see if that will work for you as an alternate? If so, please comment over there - I cannot assign you as a reviewer because GitHub - and close this.

pcm1289:
Yep, that works. Thanks, closing

PR #200, opened 2024-05-13
Use ElementTree instead of lxml

oddhack:
Alternate replacement for #198 since there's no actual need for lxml functionality or semantics in this repo.


pcm1289:
Very nice! This will definitely work as an alternative for #198

stonesthrow (approved):
Looks OK to me

oddhack:
@stonesthrow can you mark this Signed Off? It has zero header impact and you have already approved it.

PR #205, opened 2024-09-09
EGL_KHR_platform_wayland: Clarify requirements of eglSwapBuffers

jadahl:
The EGL implementation attaching the front buffer, posting damage, and committing during the call to eglSwapBuffers(WithDamage)() is a well established requirement that EGL Wayland clients currently rely upon, thus should be explicitly documented.

-----

I noticed EGL_EXT_platform_wayland already existing, but EGL_KHR_platform_wayland seems to be the one most recently updated. Should this perhaps be targeting that, and with the changes to the KHR variant being ported over?

### DIFF
--- a/extensions/KHR/EGL_KHR_platform_wayland.txt
+++ b/extensions/KHR/EGL_KHR_platform_wayland.txt
@@ -23,7 +23,7 @@ Status
 
 Version
 
-    Version 2, 2014/02/18
+    Version 4, 2024/09/09
 
 Number
 
@@ -41,7 +41,8 @@ Dependencies
     20140122).
 
     The behavior of part of this extension is different depending on whether
-    the EGL_EXT_buffer_age extension is also present.
+    the EGL_EXT_buffer_age and EGL_KHR_swap_buffers_with_damage extensions are
+    also present.
 
 Overview
 
@@ -93,6 +94,19 @@ New Behavior
     until the next buffer swap. The rationale behind this behavior is to keep
     operations result accurate until the next swap.
 
+    Add the following text to subsection 3.10.1 titled "Posting to a
+    Window" after paragraph 3 of the description of eglSwapBuffers:
+
+    "When the <surface> refers to a Wayland platform surface, synchronously
+    within the call to eglSwapBuffers the EGL implementation will invoke a
+    series of requests on the associated wl_surface, including attaching the
+    new front buffer, an appropriate surface damage via wl_surface.damage or
+    wl_surface.damage_buffer, and finally a wl_surface.commit."
+
+    If the EGL implementation associated a wp_linux_drm_syncobj_surface_v1
+    object with the surface, an acquire and a release point will be associated
+    with the surface prior to the commit as well.
+
 Issues
 
     1. Should this extension permit EGL_DEFAULT_DISPLAY as input to
@@ -107,6 +121,11 @@ Issues
        RESOLVED. No. Wayland has no pixmap type.
 
 Revision History
+    Version 4, 2024/09/09 (Jonas Ådahl)
+        - Clarify wl_surface commit requirements.
+        - Clarify wl_surface damage responsibilities.
+        - Add dependency on KHR_swap_buffers_with_damage
+
     Version 3, 2022/07/14 (Kirill Chibisov)
         - Clarify EGLSurface back buffer locking behavior with regards to
           rendering and surface querying operations.


extensions/KHR/EGL_KHR_platform_wayland.txt
```diff
@@ -93,6 +94,16 @@ New Behavior
     until the next buffer swap. The rationale behind this behavior is to keep
     operations result accurate until the next swap.
 
+    During the call to eglSwapBuffers wl_surface.attach will be requested with
```
amshafer, line 97: Should we maybe clarify that the EGL implementation is the one who will do these requests instead of just saying they "will be requested"? It seems like that's the way things like this are worded in some other extensions for clarity. 
```suggestion
    During the call to eglSwapBuffers wl_surface.attach will be requested by the EGL implementation with
```
Not sure if we also need to call out things like explicit sync points being sent by the implementation as part of this commit? I guess what you have right now just specifies the minimum set of requests the implementation will do, so that could be fine.

amshafer:
Context on the motivation for this: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/423

extensions/KHR/EGL_KHR_platform_wayland.txt
```diff
@@ -93,6 +94,16 @@ New Behavior
...
+    the new front buffer, a full surface damage is requested via
+    wl_surface.damage or wl_surface.damage_buffer, and a wl_surface.commit will
+    be requested to commit the new state.
```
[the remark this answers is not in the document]
cubanismo: To me this language already implies that, but you could say something like "Before the call to eglSwapBuffers() returns" or similar, rather than "During the call to." It could still happen in another thread, but has to happen before the call ends from the calling thread's perspective, rather than some mangled interpretation of the call's duration from the driver's point of view.

extensions/KHR/EGL_KHR_platform_wayland.txt
cubanismo: It seems hard both to enumerate all the state ever needed and to write general language that implies the right amount of state.

Maybe for explicit sync:

"if the EGL implementation associated a wp_linux_drm_syncobj_surface_v1 object with the surface, an acquire and a release point will be associated with the surface prior to the commit as well."

For others state the implementation might need to set as part of a commit, hopefully it makes more sense to call it out in separate extensions? E.g., as an interaction with this extension in a present timing extension, an HDR extension, etc., or in the other direction for existing EGL extensions that necessitate additional wayland surface state?

extensions/KHR/EGL_KHR_platform_wayland.txt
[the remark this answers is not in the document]
jadahl: Is that better? I left out mentioning threads, because I guess it's still technically possible to offload anything to a thread, as long as the thread calling eglSwapBuffers is blocked.

extensions/KHR/EGL_KHR_platform_wayland.txt
jadahl: I reworded things a bit while still trying to make it more explicit, because adding "by the EGL implementation" to every "requested" would have been a bit much.

extensions/KHR/EGL_KHR_platform_wayland.txt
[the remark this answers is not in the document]
fooishbar: I like it. Thanks!

extensions/KHR/EGL_KHR_platform_wayland.txt
jadahl: Didn't add anything about explicit sync there, as it's not strictly needed, but could be added too I guess.

extensions/KHR/EGL_KHR_platform_wayland.txt
```diff
@@ -107,6 +119,11 @@ Issues
        RESOLVED. No. Wayland has no pixmap type.
 
 Revision History
+    Version 4, 2024/09/09 (Jonas Ådahl)
```
fooishbar, line 122: Needs a bump in the header

jadahl: Fixed.

extensions/KHR/EGL_KHR_platform_wayland.txt
jadahl: > "if the EGL implementation associated a wp_linux_drm_syncobj_surface_v1 object with the surface, an acquire and a release point will be associated with the surface prior to the commit as well."

Added this as well, thanks.

cubanismo:
This version looks good to me.

emersion (approved):

emersion:
Gentle ping - is anything blocking this?

extensions/KHR/EGL_KHR_platform_wayland.txt
```diff
@@ -41,7 +41,8 @@ Dependencies
     20140122).
 
     The behavior of part of this extension is different depending on whether
-    the EGL_EXT_buffer_age extension is also present.
+    the EGL_EXT_buffer_age and KHR_swap_buffers_with_damage extensions are also
```
cubanismo, line 44: nitpick: This should be EGL_KHR_swap_buffers_with_damage.

extensions/KHR/EGL_KHR_platform_wayland.txt
```diff
@@ -93,6 +94,21 @@ New Behavior
...
+    Synchronously within the call to eglSwapBuffers the EGL implementation will
+    invoke a series of requests on the associated wl_surface, including
+    attaching the new front buffer, a full surface damage via wl_surface.damage
+    or wl_surface.damage_buffer, and finally a wl_surface.commit.
+
+    If the EGL implementation associated a wp_linux_drm_syncobj_surface_v1
+    object with the surface, an acquire and a release point will be associated
+    with the surface prior to the commit as well.
+
+    If KHR_swap_buffers_with_damage is available, the equivalent to calling
```
cubanismo, line 106: This section shouldn't really be necessary. eglSwapBuffersWithDamageKHR() already states that it inherits all the behavior of eglSwapBuffers():

    As an alternative to eglSwapBuffers use:

        EGLBoolean eglSwapBuffersWithDamageKHR (
            EGLDisplay dpy,
            EGLSurface surface,
            const EGLint *rects,
            EGLint n_rects);

        to do the same thing as eglSwapBuffers but additionally report
        a list of rectangles that define the region that has truly
        changed since the last frame.

I think after applying the above suggestion about properly inserting this language into the eglSwapBuffers() portion of the spec, this inheritance would be more naturally implied in readers' minds.

extensions/KHR/EGL_KHR_platform_wayland.txt
```diff
@@ -93,6 +94,21 @@ New Behavior
     until the next buffer swap. The rationale behind this behavior is to keep
     operations result accurate until the next swap.
 
+    Synchronously within the call to eglSwapBuffers the EGL implementation will
```
cubanismo, line 97: This doesn't really belong here IMHO. I think it should be injected into the existing definition of eglSwapBuffers() using the usual pseudo-patch semantics of GL/EGL/GLX extension specifications. E.g., "Changes to Chapter 3 of the EGL 1.5 Specification (EGL Functions and Errors)

    Add the following text to subsection 3.10.1 titled "Posting to a
    Window" after paragraph 3 of the description of eglSwapBuffers:

    "When the <surface> refers to a Wayland platform surface, <Your new language here>"

The other language in this spec should arguably be similarly inserted relative to existing spec text, but its more general in nature, so it probably isn't as important to clean that up.

cubanismo (changes requested):
Thanks for the ping. Did a more comprehensive review and suggested some formatting improvements that should make this easier to mentally parse for those familiar with the EGL spec and other extension syntax, but the guts of the proposal look good to me.

mahkoh:
Existing clients rely on EGL not creating a [wp_color_management_surface_v1](https://wayland.app/protocols/color-management-v1#wp_color_management_surface_v1). Mesa previously contained a bug where using zink caused such clients to crash because the zink WSI (which delegates to vulkan) did not ensure that this was the case.

The extension should clarify if/when the implementation creates this object.

extensions/KHR/EGL_KHR_platform_wayland.txt
jadahl: It's now "Add the following ..." for this part.

extensions/KHR/EGL_KHR_platform_wayland.txt
```diff
@@ -93,6 +94,19 @@ New Behavior
     until the next buffer swap. The rationale behind this behavior is to keep
     operations result accurate until the next swap.
 
+    Add the following text to subsection 3.10.1 titled "Posting to a
+    Window" after paragraph 3 of the description of eglSwapBuffers:
+
+    "When the <surface> refers to a Wayland platform surface, synchronously
+    within the call to eglSwapBuffers the EGL implementation will invoke a
+    series of requests on the associated wl_surface, including attaching the
+    new front buffer, a full surface damage via wl_surface.damage or
```
cubanismo: Should changes also be made to indicate less than full surface damage will be sent if the call is eglSwapBuffersWithDamageKHR? I think that command would otherwise inherit this requirement by default.

jadahl: The EXT_swap_buffers_with_damage extension adds the following to the same paragraph:

        As an alternative to eglSwapBuffers use:

        EGLBoolean eglSwapBuffersWithDamageEXT (
            EGLDisplay dpy,
            EGLSurface surface,
            EGLint *rects,
            EGLint n_rects);

        to do the same thing as eglSwapBuffers but additionally report
        a list of rectangles that define the region that has truly
        changed since the last frame. ...

so I would think we're already covered, as it already "copies" the semantics from eglSwapBuffers, adding only the damage region to it.

I'd be fine adding something though, but might end up a bit messy.

cubanismo: Yeah, but that's an additive requirement. It doesn't clearly remove the requirement to explicitly send specific Wayland protocol for full-drawable damage from eglSwapBuffers(). I think the easiest thing to do would be water-down the new language here to say something like "an appropriate damage region" instead of requiring full-surface damage.

jadahl: Good point, changed to the watered down "appropriate surface damage region".

cubanismo (approved):

stonesthrow:
I see clarification but no operational change to this spec. LGTM,

fooishbar:
lgtm, I'm happy to approve from Mesa, thanks a lot @jadahl!

fooishbar:
@stonesthrow @oddhack is anything holding up the merge of this and #222 please?

stonesthrow:
No, it has been approved, just waiting for Jon's attention.

PR #210, opened 2025-01-13
EGL_ANDROID_native_fence_sync: Fix the attribute that must be set after a flush

dkorkmazturk:
The EGL_SYNC_NATIVE_FENCE_FD_ANDROID attribute of an EGL sync object is intended to be set to a file descriptor that refers to the underlying native fence object.

However, a sentence in the specification suggests that the EGL_SYNC_NATIVE_FENCE_ANDROID attribute is used for storing file descriptors instead. Which is probably a typo since EGL_SYNC_NATIVE_FENCE_ANDROID is not an attribute.


stonesthrow:
@ianelliottus - Check this please

PR #218, opened 2025-10-30
Draft: Enforce no implicit color management on Wayland

jadahl:
EGL_KHR_platform_wayland: Enforce no color management

Wayland EGL applications rely on the EGL implementation creating
wp_color_management_surface_v1 objects for the associated surface, as
doing so may cause protocol errors when the application itself creates
them. Document this restriction.

### DIFF
--- a/extensions/KHR/EGL_KHR_platform_wayland.txt
+++ b/extensions/KHR/EGL_KHR_platform_wayland.txt
@@ -107,6 +107,9 @@ New Behavior
     object with the surface, an acquire and a release point will be associated
     with the surface prior to the commit as well.
 
+    An EGL implementation must not create a wp_color_management_surface_v1
+    object for the associated wl_surface.
+
 Issues
 
     1. Should this extension permit EGL_DEFAULT_DISPLAY as input to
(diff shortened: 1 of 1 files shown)


extensions/KHR/EGL_KHR_platform_wayland.txt
```diff
@@ -93,6 +94,22 @@ New Behavior
...
+    within the call to eglSwapBuffers the EGL implementation will invoke a
+    series of requests on the associated wl_surface, including attaching the
+    new front buffer, a full surface damage via wl_surface.damage or
+    wl_surface.damage_buffer, and finally a wl_surface.commit."
+
+    If the EGL implementation associated a wp_linux_drm_syncobj_surface_v1
+    object with the surface, an acquire and a release point will be associated
+    with the surface prior to the commit as well.
+
+    An EGL implementation must not create a wp_color_management_surface_v1
```
cubanismo: I don't think that's realistic. Don't we need this to implement things like EGL_EXT_gl_colorspace_scrgb_linear properly?

jadahl: That sounds like a necessity indeed, unless the application itself handles describing the content, and EGL just generating appropriate pixel values. Would EGL need to manage this state internally, I guess it needs to be specified when exactly it will do so, to not cause conflicts with existing color manager protocol users that uses it to describe EGL content.

cubanismo: In Vulkan we worked out a design where the PASSTHROUGH colorspace indicated the Vulkan implementation wouldn't create the Wayland object. Perhaps something similar is possible in EGL? I would expect the application isn't using a particular colorspace value currently if it's using the color management protocol directly. I haven't reviewed the EGL HDR/colorspace-related extensions in detail. I think the sRGB/linear colorspace spec has implications on how core rendering to the surface works, but I don't think the others do.

PR #223, opened 2026-04-20
EGL_EXT_yuv_surface: moot attributes have no effect

kusma:
The attributes EGL_YUV_DEPTH_RANGE_EXT and EGL_YUV_CSC_STANDARD_EXT does not affect anything that EGL would care about. These attributes instead encodes information that each client API would have to know in order to access the data in these surfaces.

...However, the only client-api specification that uses these (GL_EXT_YUV_target) leaves this up to the fragment shader, so these attributes has no effect.

There's also not anything to match against here; color-space and ranges are typically not a part of a pixel-format specification.

### DIFF
--- a/extensions/EXT/EGL_EXT_yuv_surface.txt
+++ b/extensions/EXT/EGL_EXT_yuv_surface.txt
@@ -31,7 +31,7 @@ Status
 
 Version
 
-    Version 9, May 4th, 2017
+    Version 10, May 18th, 2026
 
 Number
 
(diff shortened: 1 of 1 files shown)


stonesthrow (commented):
I agree with saying they have no effect. But I don't think we should remove any enums, they are part of API, they need to continue in the header file.

Sections: Version number and Issues will need updated.

kusma:
> I agree with saying they have no effect. But I don't think we should remove any enums, they are part of API, they need to continue in the header file.

I didn't remove any enums, but it seems like I accidentally deleted the description of the `EGL_YUV_PLANE_BPP_EXT` property, which does encode something useful. That's fixed now.

> Sections: Version number and Issues will need updated.

Updated.

stonesthrow:
@oddhack - Please look this over. I am concerned about backward compatibility. The 2 attributes identified, are extra info and not needed to describe layout for GPU. However, I don't know if any implementation uses/checks for these, and so I don't think they can be removed. They are still an attribute an implementation must check/allow.

